PR #18, opened 2025-09-06
feat(festivals): add stage order and color

chiptus:


src/lib/timelineCalculator.ts
```diff
@@ -99,16 +104,22 @@ export function calculateTimelineData(
...
-      name: stageName,
-      sets: sets.sort((a, b) => {
-        if (!a.startTime || !b.startTime) return 0;
-        return a.startTime.getTime() - b.startTime.getTime();
-      }),
-    }))
-    .sort((a, b) => a.name.localeCompare(b.name));
+  // Create unified stages array - stages will be sorted by stage_order then by name
+  const unifiedStagesUnsorted = Object.entries(allStageGroups).map(
+    ([stageName, sets]) => {
```
chiptus, line 109: should use stageId here

PR #58, opened 2026-06-25
refactor(api): sets → src/api modules (Effort 1)

chiptus:
Mechanical move of the `sets` feature from `src/hooks/queries/sets` to `src/api/sets`, adding `queryOptions` factories for the set query hooks. No behavior change. Part of #52.

## Manual verification
- In an edition, open the **Artists** and **Schedule** tabs — sets render and filter as before.
- Open a **Set details** page and cast a vote — voting + set info still work.
- In admin, open **Sets management**: create, edit, add/remove artist, delete a set — all succeed.
- `pnpm run lint && pnpm test && pnpm run build` all pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01K513rsQz6Lg1HbbfYiafrE)_


src/api/sets/useSets.ts
```diff
@@ -65,10 +44,13 @@ async function fetchSets(): Promise<FestivalSet[]> {
...
-export function useSetsQuery() {
-  return useQuery({
+export function setsQuery() {
+  return queryOptions({
     queryKey: setsKeys.lists(),
     queryFn: fetchSets,
   });
 }
+
+export function useSetsQuery() {
```
chiptus: Why do we export a hook? Is it used?

chiptus: Good catch — it wasn't used. `useSetsQuery` fetched **all** sets globally (unscoped via `setsKeys.lists()`), but every actual consumer uses the edition-scoped `useSetsByEditionQuery` instead. It had zero call sites and the module was fully orphaned — and it was already dead on `main` (this PR just mechanically carried it over and wrapped it in a `setsQuery()` factory).

Removed the whole dead chain (`fetchSets` + `setsQuery()` + `useSetsQuery`) in `d9e3d90` — deleted `src/api/sets/useSets.ts` entirely. Behavior-neutral (nothing referenced it); lint/build/290 tests still green.

---

PR #90, opened 2026-07-05
feat(router): pilot loaders + useSuspenseQuery for festival/edition/set

chiptus:
Adopts route loaders + `useSuspenseQuery` on the festival → edition → set browsing path (#53), establishing the reference pattern before the rest of the checklist rolls out.
Set detail page now loads through the router's loader with no manual loading/not-found states.

## Verification

- Visit a festival page; it loads via the route loader with no separate query round-trip.
- Navigate festival → edition → a set; the set page renders straight from the prefetched loader data, no loading flicker.
- Visit a set with a bad slug; the global error fallback renders instead of a blank page.
- `pnpm run build`, `pnpm run lint`, `pnpm test` all pass (310 tests).


src/routes/festivals/$festivalSlug/editions/$editionSlug/sets/$setSlug.tsx
```diff
@@ -1,10 +1,23 @@
...
 )({
   component: SetDetails,
   validateSearch: filterSortSearchSchema,
+  loader: async ({ params, context }) => {
+    const edition = await context.queryClient.ensureQueryData(
+      editionBySlugQuery({
+        festivalSlug: params.festivalSlug,
+        editionSlug: params.editionSlug,
+      }),
+    );
```
chiptus: shouldn't this be in the main $editionSlug route?

chiptus: Good catch — the edition was already being ensured by `editions/$editionSlug.tsx`'s `beforeLoad`, so this was a redundant re-fetch just to read `edition.id`. Fixed in 7a5bfe2: that `beforeLoad` now returns `{ edition }`, and this loader reads it off route context (`context.edition.id`) instead of re-querying.

---

src/routes/festivals/$festivalSlug/editions/$editionSlug.tsx
```diff
@@ -15,11 +15,13 @@ export const Route = createFileRoute(
       });
     }
 
-    await context.queryClient.ensureQueryData(
+    const edition = await context.queryClient.ensureQueryData(
```
chiptus, line 18: why do we need to return the value?

chiptus: This route's own component (`EditionLayout`) doesn't need it directly — it gets `edition` via `FestivalEditionContext` instead. The return is for descendants: TanStack Router merges whatever `beforeLoad` returns into `context` for that route and all its child routes/loaders. Returning `{ edition }` here is exactly what lets `sets/$setSlug.tsx`'s loader read `context.edition.id` without re-fetching (the fix from your first comment above) — without the return, that child loader would have no way to get the id except either re-querying (what we just removed) or manually reconstructing `editionsKeys.bySlug(...)` to read the cache directly, which duplicates the key logic that already lives in `editionBySlugQuery`.

---

chiptus: Agreed, you're right — threading `edition` through router context makes it a second, non-reactive copy of server state, which fights the "one queryOptions factory feeds both `ensureQueryData` and `useSuspenseQuery`" model this effort is built on. Reverted in be6e9e0: `sets/$setSlug.tsx`'s loader goes back to calling `ensureQueryData(editionBySlugQuery(...))` directly — it's a cache hit since the parent route's `beforeLoad` already prefetched it well within `staleTime`, so no extra network call, just the same factory being reused.

---

src/routes/festivals/$festivalSlug/editions/$editionSlug/sets/$setSlug.tsx
chiptus: so why do we still have it?

chiptus: Fair — we didn't actually need it. Pushed the edition lookup into `fetchSetBySlug` itself in 21bb6e1, mirroring how `fetchFestivalEditionBySlug` already resolves its festival internally rather than making callers chain it. `setBySlugQuery`/`useSetBySlugQuery` now take `{ festivalSlug, editionSlug, slug }` and resolve the edition id inside the fetch function; the route loader and `SetDetails.tsx` deal only in slugs now, no `editionBySlugQuery` import needed here at all. Confirmed `setBySlugQuery`/`setsKeys.bySlug` had no other callers, so the signature change was contained to these three files.

---

chiptus: Reverted in d90c668 — my previous fix pushed `fetchSetBySlug` to call `fetchFestivalEditionBySlug` directly, but that function does a raw (uncached) Supabase fetch, not a cache read, so it actually introduced two duplicate network round-trips (festival + edition) on every set page load, even though both were already sitting in the query cache from the parent routes' loaders. Back to: loader does `ensureQueryData(editionBySlugQuery(...))` (cache hit) → `edition.id` → `ensureQueryData(setBySlugQuery(slug, edition.id))`; `SetDetails.tsx` reads `edition.id` from `FestivalEditionContext` (already resolved there, no extra fetch). `useSetBySlugQuery`/`setsKeys.bySlug` back to `{ slug, editionId }`.

---

src/pages/SetDetails.tsx
```diff
@@ -23,25 +22,16 @@ export function SetDetails() {
   });
   const { edition, festival } = useFestivalEdition();
   const { state: urlState } = useUrlState();
-  const setQuery = useSetBySlugQuery({
-    slug: setSlug,
-    editionId: edition?.id,
-  });
+  const { data: currentSet } = useSuspenseQuery(
+    setBySlugQuery(setSlug, edition!.id),
```
chiptus: we should check whether edition is defined, no?

src/routes/festivals/$festivalSlug/editions/$editionSlug/sets/$setSlug.tsx
chiptus: maybe this should be called on the edition page? and then edition returned on the context? this way SetDetails also have it defined for sure?

chiptus: Done in c811339 — `editions/$editionSlug.tsx`'s `beforeLoad` now returns `{ edition }`, giving every route nested under it a typed non-null `edition` via router context. `sets/$setSlug.tsx`'s loader reads `context.edition.id` directly (no more `editionBySlugQuery` import/re-fetch here at all), and `SetDetails.tsx` reads it via `useRouteContext({ from: "/festivals/$festivalSlug/editions/$editionSlug" })` instead of the unsafe `edition!.id`. This is different from the router-context approach reverted earlier in this thread — that one duplicated `FestivalEditionContext`'s job; this one solves a problem `FestivalEditionContext` structurally can't (its `edition` has to stay nullable since it's shared with the edition-selection page where nothing's picked yet).

---

src/pages/SetDetails.tsx
chiptus: Fixed properly in c811339 rather than adding a guard — `edition` is now read via `useRouteContext({ from: "/festivals/$festivalSlug/editions/$editionSlug" })` instead of the nullable `FestivalEditionContext`, so it's typed non-null (guaranteed by the parent route's `beforeLoad`) and there's no `!` assertion left to double-check. See the reply on your other comment for the full reasoning.

---

src/routes/festivals/$festivalSlug/editions/$editionSlug/sets/$setSlug.tsx
chiptus: Yes, done in c811339 — `editions/$editionSlug.tsx`'s `beforeLoad` now returns `{ edition }`, so every route nested under it (and `SetDetails.tsx` via `useRouteContext`) gets a typed non-null `edition`. This is different from the router-context approach reverted earlier in this thread: that one duplicated `FestivalEditionContext`'s job for data the *component* could already get from React context. This one solves something `FestivalEditionContext` structurally can't — its `edition` has to stay nullable since it's shared with the edition-selection page where nothing's picked yet, so it can never give descendants a non-null guarantee the way router context now does.

---

src/pages/SetDetails.tsx
```diff
@@ -21,27 +20,21 @@ export function SetDetails() {
...
     from: "/festivals/$festivalSlug/editions/$editionSlug/sets/$setSlug",
   });
-  const { edition, festival } = useFestivalEdition();
-  const { state: urlState } = useUrlState();
-  const setQuery = useSetBySlugQuery({
-    slug: setSlug,
-    editionId: edition?.id,
+  const { festival } = useFestivalEdition();
+  const { edition } = useRouteContext({
+    from: "/festivals/$festivalSlug/editions/$editionSlug",
```
chiptus: Do we really need this? From? Doesn't the context inherit?

src/pages/SetDetails.tsx
chiptus: You're right, `from` is still required by the API (no way to get a typed result without it), but you're right that it doesn't need to point at the specific ancestor — context inherits down, so `sets/$setSlug`'s own merged context already has `edition`. Fixed in 727d548: `from` now matches the same leaf route already used for `useParams` right above it.

---

PR #92, opened 2026-07-05
feat(schedule): render set times in festival timezone across the app

chiptus:
Adds an optional `timezone` param to `formatTimeRange` and threads `festival.timezone` into the set-detail cards, artist SetCard metadata, and the admin SetsTable so set times render in festival time outside the Schedule tab.
Reveal-level masking (day-only vs full time) is unchanged — only the render zone moves.

Closes #77

## Verification
- Open a set-detail page with browser tz far from the festival's; start/end time and day read in festival time.
- Artist tab SetCard metadata and admin SetsTable show the same festival-time values.
- Callers that omit the new `timezone` arg are unchanged (byte-identical output); `pnpm test` covers the new tz case.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01GLgAHCzTppjx4nT8BLg3vs)_

### DIFF
--- a/src/pages/SetDetails/MultiArtistSetInfoCard.tsx
+++ b/src/pages/SetDetails/MultiArtistSetInfoCard.tsx
@@ -15,6 +15,7 @@ import { IndividualArtistCard } from "./IndividualArtistCard";
 import { StagePin } from "@/components/StagePin";
 import { MarkdownText } from "@/components/ui/markdown-text";
 import { useScheduleReveal } from "@/hooks/useScheduleReveal";
+import { useFestivalEdition } from "@/contexts/FestivalEditionContext";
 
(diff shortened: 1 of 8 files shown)


src/pages/SetDetails/MultiArtistSetInfoCard.tsx
```diff
@@ -101,6 +105,11 @@ export function MultiArtistSetInfoCard({
                     <span className="text-sm">{dayOnlyFormatted}</span>
                   </div>
                 )}
+                {(timeRangeFormatted || dayOnlyFormatted) && (
+                  <span className="w-full text-xs text-purple-200/70">
```
chiptus: this should be shared

PR #112, opened 2026-07-08
fix(festivals): add timeout and error state to festival queries

chiptus:
Adds a 10s timeout to festival/edition Supabase queries and an error state with retry, so an unreachable backend shows an actionable error instead of an infinite "Loading festivals…" spinner. Closes #56.

## Verification
- Load the festivals page normally; list renders as before.
- Simulate a hung/unreachable backend; after ~10s an error message with a retry button appears instead of spinning forever.
- Click retry; the query refetches.
- Load with zero festivals; "No Festivals Available" state is unchanged.
- Same check in admin `FestivalManagementTable`.

---

### DIFF
--- a/src/api/editions/useFestivalEditionBySlug.ts
+++ b/src/api/editions/useFestivalEditionBySlug.ts
@@ -2,28 +2,28 @@ import { queryOptions, useQuery } from "@tanstack/react-query";
 import { supabase } from "@/integrations/supabase/client";
 import { FestivalEdition, editionsKeys } from "./types";
 import { fetchFestivalBySlug } from "@/api/festivals/useFestivalBySlug";
-import { withTimeout } from "@/lib/timeout";
+import { timeoutSignal } from "@/lib/timeout";
 
 export async function fetchFestivalEditionBySlug({
   editionSlug,
   festivalSlug,
+  signal,
 }: {
   festivalSlug: string;
   editionSlug: string;
+  signal?: AbortSignal;
 }): Promise<FestivalEdition> {
-  const festival = await fetchFestivalBySlug(festivalSlug);
+  const festival = await fetchFestivalBySlug(festivalSlug, signal);
 
   try {
-    const { data, error } = await withTimeout(
-      supabase
-        .from("festival_editions")
-        .select("*")
-        .eq("archived", false)
-        .eq("festival_id", festival.id)
-        .eq("slug", editionSlug)
-        .single(),
-      10000,
-    );
+    const { data, error } = await supabase
+      .from("festival_editions")
+      .select("*")
+      .eq("archived", false)
+      .eq("festival_id", festival.id)
+      .eq("slug", editionSlug)
+      .abortSignal(timeoutSignal(signal))
+      .single();
 
     if (error) {
       throw new Error("Failed to load festival edition");
@@ -31,7 +31,7 @@ export async function fetchFestivalEditionBySlug({
 
     return data;
   } catch (err) {
-    if (err instanceof Error && err.message === "Request timeout") {
+    if (err instanceof DOMException && err.name === "TimeoutError") {
       throw new Error("Failed to load festival edition - request timed out");
     }
     throw err;
@@ -47,10 +47,11 @@ export function editionBySlugQuery({
 }) {
   return queryOptions({
     queryKey: editionsKeys.bySlug(festivalSlug, editionSlug),
-    queryFn: () =>
+    queryFn: ({ signal }) =>
       fetchFestivalEditionBySlug({
         festivalSlug,
         editionSlug,
+        signal,
       }),
   });
 }
--- a/src/api/editions/useFestivalEditionBySlug.ts
+++ b/src/api/editions/useFestivalEditionBySlug.ts
@@ -2,7 +2,7 @@ import { queryOptions, useQuery } from "@tanstack/react-query";
 import { supabase } from "@/integrations/supabase/client";
 import { FestivalEdition, editionsKeys } from "./types";
 import { fetchFestivalBySlug } from "@/api/festivals/useFestivalBySlug";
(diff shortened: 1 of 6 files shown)


src/api/editions/useFestivalEditionBySlug.ts
```diff
@@ -12,21 +13,29 @@ export async function fetchFestivalEditionBySlug({
...
   const festival = await fetchFestivalBySlug(festivalSlug);
 
-  const query = supabase
-    .from("festival_editions")
-    .select("*")
-    .eq("archived", false)
-    .eq("festival_id", festival.id)
-    .eq("slug", editionSlug)
-    .single();
+  try {
```
chiptus: no need to use a custom timeout. use 


```
const { data, error } = await supabase
  .from('large_dataset')
  .select('*')
  .abortSignal(controller.signal);
```

react query also passes an abort signal to the query function so we should merge them

chiptus: Done — dropped the custom `withTimeout` wrapper. Now merging TanStack Query's own `signal` with `AbortSignal.timeout(10000)` via `AbortSignal.any(...)` and passing that into `.abortSignal()` on the Supabase builder (pushed in be95f63).

---

src/api/editions/useFestivalEditionBySlug.ts
```diff
@@ -2,31 +2,40 @@ import { queryOptions, useQuery } from "@tanstack/react-query";
...
   editionSlug,
   festivalSlug,
+  signal,
 }: {
   festivalSlug: string;
   editionSlug: string;
+  signal?: AbortSignal;
 }): Promise<FestivalEdition> {
-  const festival = await fetchFestivalBySlug(festivalSlug);
+  const festival = await fetchFestivalBySlug(festivalSlug, signal);
```
chiptus, line 16: open a ticket to accept festival id here instead of fetching the festival obj again

src/api/editions/useFestivalEditionBySlug.ts
```diff
@@ -2,31 +2,40 @@ import { queryOptions, useQuery } from "@tanstack/react-query";
...
 
-  const query = supabase
-    .from("festival_editions")
-    .select("*")
-    .eq("archived", false)
-    .eq("festival_id", festival.id)
-    .eq("slug", editionSlug)
-    .single();
+  try {
```
chiptus: I don't think we need try catch here. supabase returns an error instead of throwing

src/api/festivals/useFestivals.ts
```diff
@@ -15,19 +17,26 @@ async function fetchFestivals({ all }: { all?: boolean } = {}): Promise<
     query = query.eq("published", true);
   }
 
-  const { data, error } = await query;
+  try {
+    const { data, error } = await query.abortSignal(timeoutSignal(signal));
```
chiptus: why do we wrap the signal here? we should probably move it to the query and accept the timeout as a parameter

src/api/editions/useFestivalEditionBySlug.ts
chiptus: Confirmed — checked `PostgrestBuilder.then()`: it wraps the fetch call in `res.catch(...)` and resolves `{ error }` for both network/abort failures unless `.throwOnError()` is set (which we don't use), so the try/catch never actually caught anything. Removed it; the timeout case is now detected via `isTimeoutError(signal)` after the awaited call resolves with an error (pushed in dd797db).

---

src/api/festivals/useFestivals.ts
chiptus: Moved it — `festivalsQuery`/`festivalBySlugQuery`/`editionBySlugQuery` now accept a `timeoutMs` parameter (default 10000) and compute the merged signal via `withTimeout(signal, timeoutMs)` in the `queryFn`, passing the already-merged signal down. The raw `fetch*` functions just take a `signal` and call `.abortSignal(signal)` — no more signal-wrapping inside them. Pushed in dd797db.

---

src/api/editions/useFestivalEditionBySlug.ts
chiptus: Opened #116 to track this.

---

src/pages/admin/festivals/FestivalManagementTable.tsx
```diff
@@ -25,7 +25,12 @@ export function FestivalManagementTable({
   onSelect: (festival: Festival) => void;
   selected: string;
 }) {
-  const { data: festivals = [], isLoading } = useFestivalsQuery({ all: true });
+  const {
+    data: festivals = [],
+    isLoading,
+    isError,
+    refetch,
+  } = useFestivalsQuery({ all: true });
```
chiptus, line 33: do we have a ticket to convert this to useSuspendQuery?

src/pages/FestivalSelection.tsx
```diff
@@ -20,8 +20,12 @@ import { TopBar } from "@/components/layout/TopBar";
...
 
 export default function FestivalSelection() {
-  const { data: availableFestivals = [], isLoading: festivalsLoading } =
-    useFestivalsQuery();
+  const {
+    data: availableFestivals = [],
+    isLoading: festivalsLoading,
+    isError: festivalsError,
+    refetch,
+  } = useFestivalsQuery();
```
chiptus, line 28: do we have a ticket to convert this to useSuspendQuery?

chiptus: Yes — #109 (candidate 3) already covers this exact site.

---

src/pages/admin/festivals/FestivalManagementTable.tsx
chiptus: #109 tracks `FestivalSelection.tsx` already but didn't mention this admin table — added it there as a second candidate site.

---

PR #113, opened 2026-07-08
fix(festivals): scope query cache keys by all filter

chiptus:
Encodes the `all` filter into festivals/editions query keys so published-only and admin (unpublished-included) variants no longer share one cache entry. Mutations still invalidate the broad prefix so both variants refresh. Closes #55.

## Verification
- Visit the public festival list, then the admin festival management table; each shows the correct filtered/unfiltered data with no cache bleed.
- Same check for edition selection vs. admin edition management.
- Create/update/delete a festival or edition; both public and admin views refetch and stay in sync.

---

### DIFF
--- a/src/api/editions/types.ts
+++ b/src/api/editions/types.ts
@@ -6,7 +6,8 @@ export type FestivalEdition =
 
 // Query key factory for festival editions
 export const editionsKeys = {
-  root: (festivalId: string) => ["festivals", festivalId, "editions"] as const,
+  root: (festivalId: string) =>
+    [...festivalsKeys.root(), festivalId, "editions"] as const,
   all: (festivalId: string, { all }: { all?: boolean } = {}) =>
     [...editionsKeys.root(festivalId), { all }] as const,
   item: ({
--- a/src/api/editions/types.ts
+++ b/src/api/editions/types.ts
@@ -1,4 +1,5 @@
 import type { Database } from "@/integrations/supabase/types";
+import { festivalsKeys } from "@/api/festivals/types";
 
 export type FestivalEdition =
   Database["public"]["Tables"]["festival_editions"]["Row"];
@@ -15,7 +16,13 @@ export const editionsKeys = {
     festivalId: string;
     editionId: string;
   }) => [...editionsKeys.root(festivalId), editionId] as const,
-  bySlugRoot: () => ["festival-editions"] as const,
   bySlug: (festivalSlug: string, editionSlug: string) =>
-    [...editionsKeys.bySlugRoot(), "slug", festivalSlug, editionSlug] as const,
+    [
+      ...festivalsKeys.root(),
+      "slug",
+      festivalSlug,
+      "editions",
+      "slug",
+      editionSlug,
+    ] as const,
 };
--- a/src/api/editions/types.ts
+++ b/src/api/editions/types.ts
@@ -6,7 +6,7 @@ export type FestivalEdition =
 // Query key factory for festival editions
 export const editionsKeys = {
   all: (festivalId: string, { all }: { all?: boolean } = {}) =>
-    ["festivals", festivalId, "editions", { all: !!all }] as const,
+    ["festivals", festivalId, "editions", { all }] as const,
   item: ({
     editionId,
     festivalId,
--- a/src/api/editions/types.ts
+++ b/src/api/editions/types.ts
@@ -5,15 +5,17 @@ export type FestivalEdition =
 
 // Query key factory for festival editions
 export const editionsKeys = {
+  root: (festivalId: string) => ["festivals", festivalId, "editions"] as const,
   all: (festivalId: string, { all }: { all?: boolean } = {}) =>
-    ["festivals", festivalId, "editions", { all }] as const,
+    [...editionsKeys.root(festivalId), { all }] as const,
   item: ({
     editionId,
     festivalId,
   }: {
     festivalId: string;
     editionId: string;
-  }) => ["festivals", festivalId, "editions", editionId] as const,
+  }) => [...editionsKeys.root(festivalId), editionId] as const,
+  bySlugRoot: () => ["festival-editions"] as const,
   bySlug: (festivalSlug: string, editionSlug: string) =>
-    ["festival-editions", "slug", festivalSlug, editionSlug] as const,
+    [...editionsKeys.bySlugRoot(), "slug", festivalSlug, editionSlug] as const,
 };
--- a/src/api/editions/types.ts
+++ b/src/api/editions/types.ts
@@ -5,7 +5,8 @@ export type FestivalEdition =
 
 // Query key factory for festival editions
 export const editionsKeys = {
-  all: (festivalId: string) => ["festivals", festivalId, "editions"] as const,
+  all: (festivalId: string, { all }: { all?: boolean } = {}) =>
+    ["festivals", festivalId, "editions", { all: !!all }] as const,
   item: ({
     editionId,
     festivalId,
--- a/src/api/editions/useCreateFestivalEdition.ts
+++ b/src/api/editions/useCreateFestivalEdition.ts
@@ -1,7 +1,6 @@
 import { useMutation, useQueryClient } from "@tanstack/react-query";
 import { useToast } from "@/hooks/use-toast";
 import { supabase } from "@/integrations/supabase/client";
-import { festivalsKeys } from "@/api/festivals/types";
 import type { Database } from "@/integrations/supabase/types";
 
 type RevealLevel = Database["public"]["Enums"]["schedule_reveal_level"];
@@ -34,7 +33,7 @@ export function useCreateFestivalEditionMutation() {
   return useMutation({
     mutationFn: createFestivalEdition,
     onSuccess: () => {
-      queryClient.invalidateQueries({ queryKey: festivalsKeys.all() });
+      queryClient.invalidateQueries({ queryKey: ["festivals"] });
       toast({
         title: "Success",
         description: "Festival edition created successfully",
--- a/src/api/editions/useCreateFestivalEdition.ts
+++ b/src/api/editions/useCreateFestivalEdition.ts
@@ -35,7 +35,7 @@ export function useCreateFestivalEditionMutation() {
     mutationFn: createFestivalEdition,
     onSuccess: (_data, variables) => {
       queryClient.invalidateQueries({
-        queryKey: editionsKeys.all(variables.festival_id),
+        queryKey: editionsKeys.root(variables.festival_id),
       });
       toast({
         title: "Success",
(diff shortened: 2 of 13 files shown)


src/api/editions/types.ts
```diff
@@ -5,7 +5,8 @@ export type FestivalEdition =
 
 // Query key factory for festival editions
 export const editionsKeys = {
-  all: (festivalId: string) => ["festivals", festivalId, "editions"] as const,
+  all: (festivalId: string, { all }: { all?: boolean } = {}) =>
+    ["festivals", festivalId, "editions", { all: !!all }] as const,
```
chiptus: why isn't this enough?

```suggestion
    ["festivals", festivalId, "editions", { all }] as const,
```

src/api/festivals/types.ts
```diff
@@ -4,7 +4,8 @@ export type Festival = Database["public"]["Tables"]["festivals"]["Row"];
 
 // Query key factory for festivals
 export const festivalsKeys = {
-  all: () => ["festivals"] as const,
+  all: ({ all }: { all?: boolean } = {}) =>
+    ["festivals", { all: !!all }] as const,
```
chiptus:
```suggestion
    ["festivals", { all }] as const,
```

src/api/festivals/useCreateFestival.ts
```diff
@@ -28,7 +27,7 @@ export function useCreateFestivalMutation() {
   return useMutation({
     mutationFn: createFestival,
     onSuccess: () => {
-      queryClient.invalidateQueries({ queryKey: festivalsKeys.all() });
+      queryClient.invalidateQueries({ queryKey: ["festivals"] });
```
chiptus: why? revert

src/api/editions/types.ts
chiptus: You're right, fixed — real call sites never pass an explicit `false`, so the `!!` normalization wasn't buying anything. Pushed in d317a4a.

---

src/api/festivals/types.ts
chiptus: Same fix applied here. Pushed in d317a4a.

---

src/api/festivals/useCreateFestival.ts
chiptus: Reverted this one to `festivalsKeys.all()` (and did the same for `useCreateFestivalEdition.ts` → `editionsKeys.all(variables.festival_id)`) — a brand-new resource can't have a stale `item()`/`bySlug()` cache entry, so the narrower key is sufficient there.

Left the broad `["festivals"]` prefix as-is on update/delete though, since those mutations touch an *existing* resource whose `item()`/`bySlug()` cache does need busting (e.g. renaming a festival's slug), which `festivalsKeys.all()` alone wouldn't invalidate. Let me know if you'd rather those also go through a named key (e.g. a `festivalsKeys.root()`) instead of the bare literal — happy to add that if you'd prefer avoiding the magic string there too.

Pushed in d317a4a.

---

src/api/editions/useDeleteFestivalEdition.ts
```diff
@@ -20,7 +19,7 @@ export function useDeleteFestivalEditionMutation() {
   return useMutation({
     mutationFn: deleteFestivalEdition,
     onSuccess: () => {
-      queryClient.invalidateQueries({ queryKey: festivalsKeys.all() });
+      queryClient.invalidateQueries({ queryKey: ["festivals"] });
```
chiptus: why can't we use festivaKeys.all() ?

chiptus: Same reasoning as the create-mutation thread, but flipped: delete (and update) touch an *existing* edition, which may already have cached `item()`/`bySlug()` entries (e.g. someone has the edition detail page open) that also need busting — `editionsKeys.all(festivalId)` alone wouldn't invalidate those, only the list-query variants.

There's also a practical wrinkle here specifically: `deleteFestivalEdition`'s `mutationFn` only takes `editionId` (not `festivalId`), so `editionsKeys.all(festivalId, ...)` isn't available in `onSuccess` without either threading `festivalId` through the mutation variables or looking it up from the deleted edition's cached data first.

Happy to narrow this if you'd rather — plumbing `festivalId` through and invalidating `editionsKeys.all(festivalId)` alongside `editionsKeys.item(...)`/`bySlug(...)` explicitly would be more precise than the blanket `["festivals"]`. Let me know if you want that instead, otherwise leaving as-is.

---

src/api/festivals/useCreateFestival.ts
chiptus: yes, don't use magic string. use .root, and make all the other query keys call the root one ( item(id): ...queryKeys.root(), ...]

chiptus: Added `festivalsKeys.root()` / `editionsKeys.root(festivalId)` — `all()`/`item()`/`bySlug()` now build off them (`item(id) => [...root(), id]` etc, matching your example), and update/delete mutations invalidate via `root()` instead of the literal.

For editions specifically, `root()` needs `festivalId`, which the update/delete mutations didn't have in their variables before — threaded it through (`ScheduleRevealControl` → `FestivalEdition.tsx` now passes `festivalId={currentEdition.festival_id}`, and `FestivalEditionManagement.tsx`'s delete call now passes it too). Bonus fix: this also makes edition invalidation correctly scoped to *one* festival's editions instead of matching every festival's editions via the old bare `["festivals"]`/`["festival-editions"]` literals.

Pushed in a27426b.

---

src/api/editions/useDeleteFestivalEdition.ts
chiptus: Resolved via the `root()` refactor — see the reply on the create-mutation thread. `festivalId` is now threaded through this mutation's variables and invalidation uses `editionsKeys.root(festivalId)`. Pushed in a27426b.

---

src/api/editions/useDeleteFestivalEdition.ts
```diff
@@ -18,9 +18,13 @@ export function useDeleteFestivalEditionMutation() {
...
-    mutationFn: deleteFestivalEdition,
-    onSuccess: () => {
-      queryClient.invalidateQueries({ queryKey: festivalsKeys.all() });
+    mutationFn: ({ editionId }: { editionId: string; festivalId: string }) =>
+      deleteFestivalEdition(editionId),
+    onSuccess: (_data, variables) => {
+      queryClient.invalidateQueries({
+        queryKey: editionsKeys.root(variables.festivalId),
+      });
+      queryClient.invalidateQueries({ queryKey: editionsKeys.bySlugRoot() });
```
chiptus: why by slug instead of `festivalsKeys.root()`

src/api/editions/useDeleteFestivalEdition.ts
chiptus: Because `editionsKeys.bySlug` currently lives under a *different* top-level segment — `["festival-editions", "slug", festivalSlug, editionSlug]` (pre-existing, not something this PR introduced) — vs. everything else which nests under `["festivals", ...]`. So `festivalsKeys.root()` (`["festivals"]`) doesn't prefix-match it; that's why it needs its own invalidation call today.

I could nest `bySlug` under `"festivals"` too (e.g. `["festivals", "slug", festivalSlug, "editions", "slug", editionSlug]`) so one `festivalsKeys.root()` call covers everything and both invalidation lines here collapse into one. The tradeoff: that widens invalidation back to *all* festivals on every edition update/delete, undoing the narrow `editionsKeys.root(festivalId)` scoping we just added (which was the actual point of the `festivalId`-threading). Since we don't have `festivalSlug` in these mutations (only `festivalId`), there's no way to scope a bySlug invalidation to just the one festival without also passing the slug through.

Want me to unify it under `festivalsKeys.root()` and accept the broader invalidation, or keep the narrow id-tree scoping and leave `bySlug` as its own (broader-but-isolated) invalidation like today?

---

src/api/editions/useCreateFestivalEdition.ts
```diff
@@ -33,8 +33,10 @@ export function useCreateFestivalEditionMutation() {
 
   return useMutation({
     mutationFn: createFestivalEdition,
-    onSuccess: () => {
-      queryClient.invalidateQueries({ queryKey: festivalsKeys.all() });
+    onSuccess: (_data, variables) => {
+      queryClient.invalidateQueries({
+        queryKey: editionsKeys.all(variables.festival_id),
+      });
```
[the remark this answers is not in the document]
chiptus: Confirmed and fixed — `editionsKeys.all(festivalId)`'s `{ all: undefined }` doesn't partial-match the cached `{ all: true }` admin variant. Switched to `editionsKeys.root(festivalId)`. Pushed in e1a796c.

---

src/api/festivals/types.ts
```diff
@@ -4,8 +4,10 @@ export type Festival = Database["public"]["Tables"]["festivals"]["Row"];
 
 // Query key factory for festivals
 export const festivalsKeys = {
-  all: () => ["festivals"] as const,
-  item: (festivalId: string) => ["festivals", festivalId] as const,
+  root: () => ["festivals"] as const,
+  all: ({ all }: { all?: boolean } = {}) =>
+    [...festivalsKeys.root(), { all }] as const,
+  item: (festivalId: string) => [...festivalsKeys.root(), festivalId] as const,
```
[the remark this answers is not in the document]
chiptus: Confirmed with a quick standalone repro of TanStack Query's partial-match logic — `["festivals", { all: undefined }]` does not match a cache entry keyed `["festivals", { all: true }]`. Switched `useCreateFestivalMutation` to `festivalsKeys.root()`. Pushed in e1a796c.

---

src/api/editions/useDeleteFestivalEdition.ts
chiptus: Done — nested `editionsKeys.bySlug` under `festivalsKeys.root()` (was a separate `"festival-editions"` top-level segment), so a single `festivalsKeys.root()` call now covers festivals + editions by id + editions by slug. Dropped the second invalidation call and, since `festivalId` is no longer needed by these mutations, reverted the threading through `ScheduleRevealControl`/`FestivalEdition.tsx`/`FestivalEditionManagement.tsx` back to what it was before. Pushed in fc95fc9.

---

src/api/editions/types.ts
```diff
@@ -1,18 +1,28 @@
 import type { Database } from "@/integrations/supabase/types";
+import { festivalsKeys } from "@/api/festivals/types";
 
 export type FestivalEdition =
   Database["public"]["Tables"]["festival_editions"]["Row"];
 
 // Query key factory for festival editions
 export const editionsKeys = {
-  all: (festivalId: string) => ["festivals", festivalId, "editions"] as const,
+  root: (festivalId: string) => ["festivals", festivalId, "editions"] as const,
```
chiptus:
```suggestion
  root: (festivalId: string) => [...festivalKeys.root(), "editions"] as const,
```

chiptus: Applied with `festivalId` kept in the key (your suggestion dropped it, which would collapse every festival's editions into the same `["festivals", "editions"]` cache entry, breaking the per-festival scoping):

```ts
root: (festivalId: string) =>
  [...festivalsKeys.root(), festivalId, "editions"] as const,
```

Pushed in dd39fc9.

---

PR #115, opened 2026-07-08
fix(filters): coerce query-string search schema fields

chiptus:
Uses `z.coerce.number()` for `minRating` and explicit true/false enum parsing for `use24Hour`/`sortLocked` so these filter values round-trip correctly through URL query strings. Closes #80.

## Verification
- Load `.../sets?minRating=3`; rating filter shows 3, not the default 0.
- Toggle 24-hour time off; URL gets `use24Hour=false` and the setting stays off after reload.
- Toggle sort-lock on/off; URL param round-trips correctly on reload.
- Navigate away and back; all three params persist.

---

### DIFF
--- a/src/lib/searchSchemas.ts
+++ b/src/lib/searchSchemas.ts
@@ -14,12 +14,18 @@ export const filterSortSearchSchema = z.object({
   sort: sortOptionSchema.catch("popularity-desc"),
   stages: z.array(z.string()).catch([]),
   genres: z.array(z.string()).catch([]),
-  minRating: z.number().catch(0),
+  minRating: z.coerce.number().catch(0),
   timelineView: timelineViewSchema.catch("list"),
-  use24Hour: z.boolean().catch(true),
+  use24Hour: z
+    .enum(["true", "false"])
+    .catch("true")
+    .transform((v) => v === "true"),
   groupId: z.string().optional(),
   invite: z.string().optional(),
-  sortLocked: z.boolean().catch(false),
+  sortLocked: z
+    .enum(["true", "false"])
+    .catch("false")
+    .transform((v) => v === "true"),
   votePerspective: z.string().optional(),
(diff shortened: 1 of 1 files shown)


src/lib/searchSchemas.ts
```diff
@@ -14,12 +14,18 @@ export const filterSortSearchSchema = z.object({
   sort: sortOptionSchema.catch("popularity-desc"),
   stages: z.array(z.string()).catch([]),
   genres: z.array(z.string()).catch([]),
-  minRating: z.number().catch(0),
+  minRating: z.coerce.number().catch(0),
   timelineView: timelineViewSchema.catch("list"),
-  use24Hour: z.boolean().catch(true),
+  use24Hour: z
```
chiptus: why not z.coerce.boolean?

chiptus: `z.coerce.boolean()` coerces via JS truthiness, so any non-empty string is `true` — `?use24Hour=false` would incorrectly parse to `true` (the string `"false"` is truthy). Used `z.enum(["true","false"]).transform(v => v === "true")` instead so `false` actually round-trips, since `FilterSortControls.tsx` does set these params to `false` via the URL.

---

src/lib/searchSchemas.ts
```diff
@@ -14,12 +14,18 @@ export const filterSortSearchSchema = z.object({
...
+    .enum(["true", "false"])
+    .catch("true")
+    .transform((v) => v === "true"),
   groupId: z.string().optional(),
   invite: z.string().optional(),
-  sortLocked: z.boolean().catch(false),
+  sortLocked: z
+    .enum(["true", "false"])
+    .catch("false")
+    .transform((v) => v === "true"),
```
chiptus: create a shared "booleanStringSchema"

PR #129, opened 2026-07-10
fix(artists): add updated_at trigger

chiptus:
Adds a DB trigger so `artists.updated_at` is set on every UPDATE, matching the pattern already used by `sets`, `stages`, `groups`, etc. Removes the now-redundant manual `updated_at` assignment in `useUpdateArtist`.

## Verification
- Edit an artist's name via the admin UI and confirm `updated_at` changes in Supabase without the client setting it.
- Run `pnpm test` — 311 tests pass.
- Run `pnpm run lint` — 0 warnings/errors.

Closes #38

---
_Generated by [Claude Code](https://claude.ai/code/session_01GCbsLsv1a5jNZQhw3GnD8h)_

### DIFF
--- a/src/api/artists/useUpdateArtist.ts
+++ b/src/api/artists/useUpdateArtist.ts
@@ -33,7 +33,6 @@ async function updateArtist(variables: {
   const { genre_ids } = updates;
 
   const updateData: ArtistUpdate = {
-    updated_at: new Date().toISOString(),
     name: updates.name,
     description: updates.description,
     estimated_date: updates.estimated_date,
--- a/src/api/artists/useUpdateArtist.ts
+++ b/src/api/artists/useUpdateArtist.ts
@@ -32,33 +32,52 @@ async function updateArtist(variables: {
   const { id, updates } = variables;
   const { genre_ids } = updates;
 
-  const updateData: ArtistUpdate = {
-    name: updates.name,
-    description: updates.description,
-    estimated_date: updates.estimated_date,
-    image_url: updates.image_url,
-    soundcloud_url: updates.soundcloud_url,
-    spotify_url: updates.spotify_url,
-    stage: updates.stage,
-    time_start: updates.time_start,
-    time_end: updates.time_end,
-    archived: updates.archived,
-  };
+  const updateData: ArtistUpdate = {};
   if (updates.name !== undefined) {
+    updateData.name = updates.name;
     updateData.slug = generateSlug(updates.name);
   }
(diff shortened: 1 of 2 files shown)


src/api/artists/useUpdateArtist.ts
```diff
@@ -33,7 +33,6 @@ async function updateArtist(variables: {
   const { genre_ids } = updates;
 
   const updateData: ArtistUpdate = {
-    updated_at: new Date().toISOString(),
     name: updates.name,
     description: updates.description,
     estimated_date: updates.estimated_date,
```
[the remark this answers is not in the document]
chiptus: Good catch — confirmed `BulkEditorTableRow`'s genre column hits exactly this path. Fixed in 8bbdc1d by building `updateData` from only the provided fields and falling back to a plain `select` when there's nothing to update, so genre-only edits no longer send an empty `.update()` body.

---

src/api/artists/useUpdateArtist.ts
```diff
@@ -32,34 +32,52 @@ async function updateArtist(variables: {
   const { id, updates } = variables;
   const { genre_ids } = updates;
 
-  const updateData: ArtistUpdate = {
```
chiptus, line 35: is this change more correct? won't just removing ` updated_at: new Date().toISOString(),` a cleaner change? don't make fixes, just tell me why do it in this way

chiptus: No — plain removal of that one line isn't equivalent, it's actually what caused the bug Copilot found above.

Before the fix, the object was built unconditionally with every field listed (`name: updates.name, description: updates.description, ...`). When only `genre_ids` is passed (the bulk-editor genre column does exactly this), every other field is `undefined`. `JSON.stringify` drops `undefined`-valued keys, so `.update(updateData)` serialized to an empty `{}` body, which PostgREST rejects (no columns in `SET`). That was previously masked because `updated_at: new Date().toISOString()` was always present, guaranteeing at least one real key — so "just delete that one line" reproduces the exact bug, it doesn't fix it.

The current version does two things instead: (1) builds `updateData` by only assigning keys that were explicitly provided, so it naturally has zero keys when there's nothing to update, rather than many keys all set to `undefined`; (2) when there are zero keys, skips `.update()` entirely and does a plain `.select()` instead, since there's nothing to write.

---

chiptus: do we really need update to return an object?

PR #142, opened 2026-07-10
chore(ci): pin supabase cli version explicitly

chiptus:
Bumps the pinned Supabase CLI version to 2.109.1 in `package.json` and the two deploy workflows, since the previous `^2.58.5` pin combined with the action's auto-detect resolved unreliably (observed resolving `v2.20.3` instead).

## Verification
- Run the `db_migrate` workflow via `workflow_dispatch` on staging; confirm `supabase/setup-cli` resolves `2.109.1`.
- Run the `deploy_edge_functions` workflow via `workflow_dispatch`; confirm functions deploy successfully.

Closes #33

---
_Generated by [Claude Code](https://claude.ai/code/session_01WnMoytqBEnK3TCeRZvW2yE)_

### DIFF
--- a/.github/workflows/_db_migrate.yml
+++ b/.github/workflows/_db_migrate.yml
@@ -28,7 +28,9 @@ jobs:
 
       - uses: actions/checkout@v4
 
-      - uses: supabase/setup-cli@v3
+      - uses: supabase/setup-cli@v2
+        with:
+          version: 2.109.1
 
       - name: Push migrations
         env:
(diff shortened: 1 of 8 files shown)


.github/workflows/_db_migrate.yml
```diff
@@ -29,6 +29,8 @@ jobs:
       - uses: actions/checkout@v4
 
       - uses: supabase/setup-cli@v2
```
chiptus: try to use v3. maybe in that it will use the lock file version like https://github.com/supabase/setup-cli is saying

package.json
```diff
@@ -125,7 +125,7 @@
     "oxlint": "^1.11.1",
     "postcss": "^8.4.47",
     "prettier": "^3.6.2",
-    "supabase": "^2.58.5",
+    "supabase": "2.58.5",
```
chiptus: after tests succeed, let's try to upgrade to a fresher version, but not the latest

PR #144, opened 2026-07-10
fix(editions): accept festival id in edition-by-slug fetch

chiptus:
Threads an optional `festivalId` through `fetchFestivalEditionBySlug` so callers that already have the festival id (route loaders, `FestivalEditionProvider`) skip a redundant `fetchFestivalBySlug` call. Falls back to deriving it from the slug when unavailable. The query key is unchanged, so slug-only and slug+id calls still share one cache entry.

## Verification
- Load a festival edition page as a normal user; edition loads normally with one fewer network request for the festival lookup.
- Load an edition page directly by URL with an empty cache; edition still loads correctly via the slug fallback.
- Load the admin Set Management page for an edition; sets and edition data load correctly.

Closes #116

---
_Generated by [Claude Code](https://claude.ai/code/session_01WnMoytqBEnK3TCeRZvW2yE)_

### DIFF
--- a/src/api/editions/useFestivalEditionBySlug.ts
+++ b/src/api/editions/useFestivalEditionBySlug.ts
@@ -1,28 +1,22 @@
 import { queryOptions, useQuery } from "@tanstack/react-query";
 import { supabase } from "@/integrations/supabase/client";
 import { FestivalEdition, editionsKeys } from "./types";
-import { fetchFestivalBySlug } from "@/api/festivals/useFestivalBySlug";
 import { isTimeoutError, withTimeout } from "@/lib/timeout";
 
 export async function fetchFestivalEditionBySlug({
   editionSlug,
-  festivalSlug,
   festivalId,
   signal,
 }: {
-  festivalSlug: string;
   editionSlug: string;
-  festivalId?: string;
+  festivalId: string;
   signal?: AbortSignal;
 }): Promise<FestivalEdition> {
-  const resolvedFestivalId =
-    festivalId ?? (await fetchFestivalBySlug(festivalSlug, signal)).id;
-
   const { data, error } = await supabase
     .from("festival_editions")
     .select("*")
     .eq("archived", false)
-    .eq("festival_id", resolvedFestivalId)
+    .eq("festival_id", festivalId)
     .eq("slug", editionSlug)
     .abortSignal(signal!)
     .single();
@@ -39,22 +33,19 @@ export async function fetchFestivalEditionBySlug({
 
 export function editionBySlugQuery({
   editionSlug,
-  festivalSlug,
   festivalId,
   timeoutMs = 10000,
 }: {
-  festivalSlug: string;
   editionSlug: string;
-  festivalId?: string;
+  festivalId: string;
   timeoutMs?: number;
 }) {
   return queryOptions({
-    queryKey: editionsKeys.bySlug(festivalSlug, editionSlug),
+    queryKey: editionsKeys.bySlug(festivalId, editionSlug),
     queryFn: ({ signal }) =>
       fetchFestivalEditionBySlug({
-        festivalSlug,
-        editionSlug,
         festivalId,
+        editionSlug,
         signal: withTimeout(signal, timeoutMs),
       }),
   });
@@ -62,19 +53,16 @@ export function editionBySlugQuery({
 
 export function useFestivalEditionBySlugQuery({
   editionSlug,
-  festivalSlug,
   festivalId,
 }: {
-  festivalSlug?: string;
   editionSlug?: string;
   festivalId?: string;
 }) {
   return useQuery({
     ...editionBySlugQuery({
-      festivalSlug: festivalSlug!,
+      festivalId: festivalId!,
       editionSlug: editionSlug!,
-      festivalId,
     }),
-    enabled: !!festivalSlug && !!editionSlug,
+    enabled: !!festivalId && !!editionSlug,
   });
 }
--- a/src/api/editions/useFestivalEditionBySlug.ts
+++ b/src/api/editions/useFestivalEditionBySlug.ts
@@ -7,19 +7,22 @@ import { isTimeoutError, withTimeout } from "@/lib/timeout";
 export async function fetchFestivalEditionBySlug({
   editionSlug,
   festivalSlug,
+  festivalId,
   signal,
 }: {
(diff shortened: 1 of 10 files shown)


src/api/editions/useFestivalEditionBySlug.ts
```diff
@@ -7,19 +7,22 @@ import { isTimeoutError, withTimeout } from "@/lib/timeout";
...
+  festivalId,
   signal,
 }: {
   festivalSlug: string;
   editionSlug: string;
+  festivalId?: string;
   signal?: AbortSignal;
 }): Promise<FestivalEdition> {
-  const festival = await fetchFestivalBySlug(festivalSlug, signal);
+  const resolvedFestivalId =
```
chiptus: this means we require festivalSlug anyway, which makes this PR a bit redundant. remove festivalSlug

src/pages/admin/festivals/SetsManagement/SetManagement.tsx
```diff
@@ -15,11 +15,12 @@ export function SetManagement() {
   const { festivalSlug, editionSlug } = useParams({
     from: "/admin/festivals/$festivalSlug/editions/$editionSlug/sets",
   });
+  const festivalQuery = useFestivalBySlugQuery(festivalSlug);
   const editionQuery = useFestivalEditionBySlugQuery({
     festivalSlug,
     editionSlug,
+    festivalId: festivalQuery.data?.id,
   });
```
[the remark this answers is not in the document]
chiptus: Fixed in a6c4109 — switched to `useSuspenseQuery(festivalBySlugQuery(festivalSlug))` for the festival, so `festivalId` is guaranteed resolved before the edition query fires (no more relying on `enabled` gating).

---

src/api/editions/useFestivalEditionBySlug.ts
chiptus: Fair — checked the router source and the 3 route-loader call sites (`beforeLoad` opportunistically reading `getQueryData`) never actually worked: TanStack Router runs every matched route's `beforeLoad` to completion before any `loader` starts, so a child's `beforeLoad` can never see data a parent's `loader` fetched via `ensureQueryData`. Only cold cache/direct navigation exercises that, which is exactly the case that matters, so it was dead code.

Reverted those 3 to plain slug-only fetches in a6c4109, keeping `festivalId` only where it's a genuine win: `FestivalEditionContext` (festival arrives as a prop from a `useSuspenseQuery`-backed parent, so it's always synchronously available) and `SetManagement` (now also on `useSuspenseQuery`, see other thread).

---

chiptus: Correction to my last reply — you were right to push on this. `FestivalEdition.tsx` and `StageManagement.tsx` *do* have festival data available: both render inside `FestivalDetail`'s `<Outlet/>`, and `FestivalDetail` already resolves+gates on the festival query before ever rendering its children. That's a structural guarantee via React's mount order (parent always mounts and resolves before the nested route renders), not a timing coincidence like the router-loader case. Added `useSuspenseQuery(festivalBySlugQuery(festivalSlug))` to both in 17788bc, so `festivalId` is now threaded through every real call site except the 3 route loaders (which genuinely can't get it before their `beforeLoad` runs).

---

src/routes/admin/festivals/$festivalSlug/editions/$editionSlug/import.tsx
```diff
@@ -1,18 +1,25 @@
...
 
 export const Route = createFileRoute(
   "/admin/festivals/$festivalSlug/editions/$editionSlug/import",
 )({
-  loader: ({ params, context }) =>
-    context.queryClient.ensureQueryData(
+  loader: async ({ params, context }) => {
+    const festival = await context.queryClient.ensureQueryData(
+      festivalBySlugQuery(params.festivalSlug),
+    );
```
chiptus: don't we have the festival on the context? from ""/admin/festivals/$festivalSlug"?

src/routes/admin/festivals/$festivalSlug/editions/$editionSlug.tsx
```diff
@@ -15,9 +16,12 @@ export const Route = createFileRoute(
       });
     }
 
+    const festival = await context.queryClient.ensureQueryData(
+      festivalBySlugQuery(params.festivalSlug),
+    );
```
chiptus: same

src/routes/admin/festivals/$festivalSlug/editions/$editionSlug/import.tsx
chiptus: do we really need to warm the cache here? don't we have it fromm src/routes/admin/festivals/$festivalSlug/editions/$editionSlug.tsx?

chiptus: You're right — fixed in e7d1c5d. `/admin/festivals/$festivalSlug` now resolves the festival in `beforeLoad` and returns it via route context, so both this loader and `$editionSlug.tsx`'s `beforeLoad` read `context.festival` instead of each re-fetching it.

---

src/routes/admin/festivals/$festivalSlug/editions/$editionSlug.tsx
chiptus: Fixed in e7d1c5d, same change — `context.festival` now comes from the parent route's `beforeLoad`.

---

src/routes/admin/festivals/$festivalSlug/editions/$editionSlug/import.tsx
```diff
@@ -9,22 +8,21 @@ export const Route = createFileRoute(
   loader: ({ params, context }) =>
     context.queryClient.ensureQueryData(
       editionBySlugQuery({
```
chiptus: does import really need to warm the edition cache? isn't it available from src/routes/admin/festivals/$festivalSlug/editions/$editionSlug.tsx

chiptus: Right, no need — fixed in 79bd771. The loader is gone; the component now reads `context.edition` (and `context.festival`) directly via `Route.useRouteContext()` since `$editionSlug.tsx`'s `beforeLoad` already resolved and cached it.

---

PR #145, opened 2026-07-10
feat(phases): pre-schedule/planning edition UX

chiptus:
Adds the Pre-Schedule/Planning phase banner, a phase-driven default landing tab, and phase-aware "Lineup"/"Vote" primary-tab labeling, per #137.
No tab is ever added or hidden — only the default route, label, and banner copy change by phase.

## Verification
- Visit an edition with `schedule_reveal_level = "draft"`: lands on the sets tab, tab reads "Lineup", banner explains the schedule is coming soon.
- Visit an edition past `draft` but before the Live window: lands on the sets tab, tab reads "Vote", banner shows a countdown of days to `start_date`.
- Confirm the Schedule/Info/Map/etc. tabs are still all present and reachable in both phases.
- Visit with a `start_date` of null: banner/tab still render sensibly without throwing.

Closes #137.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FUwNzs67viT4UnYJwqjSWo)_

### DIFF
--- a/src/pages/EditionView/PhaseBanner.tsx
+++ b/src/pages/EditionView/PhaseBanner.tsx
@@ -27,7 +27,7 @@ function bannerMessage(
   timezone: string,
 ): string | null {
   if (phase === "pre-schedule") {
-    return "The lineup is live — schedule coming soon. Vote for who you want to see!";
+    return "Artists are being announced — schedule coming soon. Vote for who you want to see!";
   }
 
   if (phase === "planning") {
--- a/src/pages/EditionView/PhaseBanner.tsx
+++ b/src/pages/EditionView/PhaseBanner.tsx
@@ -1,0 +1,40 @@
+import { useFestivalPhase } from "@/hooks/useFestivalPhase";
+import { useFestivalEdition } from "@/contexts/FestivalEditionContext";
+import { daysUntilStart } from "@/lib/festivalCountdown";
+import type { FestivalPhase } from "@/lib/festivalPhase";
+
+export function PhaseBanner() {
+  const { phase } = useFestivalPhase();
+  const { edition } = useFestivalEdition();
+
+  const message = bannerMessage(phase, edition?.start_date ?? null);
+  if (!message) return null;
+
+  return (
+    <div className="mb-4 rounded-lg bg-white/10 backdrop-blur-md px-4 py-2 text-center text-sm text-purple-100">
+      {message}
+    </div>
+  );
+}
+
+function bannerMessage(
+  phase: FestivalPhase,
+  startDate: string | null,
+): string | null {
+  if (phase === "pre-schedule") {
+    return "The lineup is live — schedule coming soon. Vote for who you want to see!";
+  }
+
+  if (phase === "planning") {
+    return planningCopy(daysUntilStart(startDate, new Date()));
+  }
+
+  return null;
+}
+
+function planningCopy(days: number | null): string {
+  if (days === null) return "Schedule is out — start planning your visit!";
+  if (days <= 0) return "Schedule is out — the festival is here!";
+  if (days === 1) return "Schedule is out — 1 day to go!";
+  return `Schedule is out — ${days} days to go!`;
+}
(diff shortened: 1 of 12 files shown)


src/pages/EditionView/TabNavigation/defaultTab.ts
```diff
@@ -0,0 +1,13 @@
+import type { FestivalPhase } from "@/lib/festivalPhase";
+import type { MainTab } from "./types";
+
+const DEFAULT_TAB_BY_PHASE: Record<FestivalPhase, MainTab> = {
+  "pre-schedule": "sets",
+  planning: "sets",
+  live: "sets",
```
chiptus: I think that in live it should be timeline, no?

src/pages/EditionView/PhaseBanner.tsx
```diff
@@ -0,0 +1,45 @@
...
+  );
+}
+
+function bannerMessage(
+  phase: FestivalPhase,
+  startDate: string | null,
+  timezone: string,
+): string | null {
+  if (phase === "pre-schedule") {
+    return "The lineup is live — schedule coming soon. Vote for who you want to see!";
```
chiptus: what if only part of the lineup is live?

src/routes/festivals/$festivalSlug/editions/$editionSlug.tsx
```diff
@@ -2,15 +2,39 @@ import { createFileRoute, redirect } from "@tanstack/react-router";
...
 
 export const Route = createFileRoute(
   "/festivals/$festivalSlug/editions/$editionSlug",
 )({
   component: EditionLayout,
   beforeLoad: async ({ params, location, context }) => {
     if (params?.editionSlug && location.pathname.endsWith(params.editionSlug)) {
+      const [festival, edition] = await Promise.all([
+        context.queryClient.ensureQueryData(
+          festivalBySlugQuery(params.festivalSlug),
```
chiptus: don't we have the festival on ht econtext?

chiptus: and are we loading the edition twice? (l43)

src/pages/EditionView/TabNavigation/defaultTab.ts
chiptus: Agreed — pushed: Live now defaults to `schedule` (which itself redirects to `timeline`), instead of `sets`.

---

src/routes/festivals/$festivalSlug/editions/$editionSlug.tsx
chiptus: Good catch — we did already load it in the parent route's `loader`, but that ran too late: TanStack Router runs `beforeLoad` top-down across the whole matched tree before any `loader` runs, so this child's `beforeLoad` was hitting the network before the parent had a chance to populate the cache. Moved the festival fetch into the parent's `beforeLoad` and returned it via router context, so this route now reads `context.festival` directly instead of re-fetching.

---

chiptus: Fixed — the edition is now fetched once at the top of `beforeLoad` and reused for both the redirect-phase computation and the fallthrough return, instead of being fetched again on line 43.

---

src/pages/EditionView/PhaseBanner.tsx
chiptus: Fair point — "the lineup is live" implies completeness we can't back, since artists get announced incrementally before the schedule drops. Softened to "Artists are being announced — schedule coming soon. Vote for who you want to see!", which doesn't claim the full lineup is out. Happy to adjust the wording further if you had something more specific in mind.

---

src/routes/festivals/$festivalSlug/editions/$editionSlug.tsx
```diff
@@ -2,27 +2,38 @@ import { createFileRoute, redirect } from "@tanstack/react-router";
...
-    }
-
     const edition = await context.queryClient.ensureQueryData(
       editionBySlugQuery({
         festivalSlug: params.festivalSlug,
         editionSlug: params.editionSlug,
       }),
     );
 
+    if (params?.editionSlug && location.pathname.endsWith(params.editionSlug)) {
```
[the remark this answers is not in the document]
chiptus: Good catch, fixed — now compares `location.pathname` against the exact base path (`/festivals/$festivalSlug/editions/$editionSlug`, with-or-without trailing slash) instead of `endsWith`.

---

PR #146, opened 2026-07-10
feat(phases): add override + set_ratings table

chiptus:
Adds the epic's one migration — nullable `phase_override` on the edition plus `set_ratings` (RLS mirroring `votes`) — and the `getEffectiveFestivalPhase` seam (`override ?? derived`), with a Core Team admin control to set/clear the override. Per #139.

Note: this PR does not by itself change how the edition's default-tab redirect picks a phase (that logic lives in #137's branch). Once both #137 and #139 are merged, `src/routes/festivals/$festivalSlug/editions/$editionSlug.tsx`'s `beforeLoad` should be updated to call `getEffectiveFestivalPhase` instead of `getFestivalPhase` so an admin override also affects the landing tab, not just the banner/label — flagging this as a small follow-up once both land.

## Verification
- In the edition admin area, set a phase override (e.g. force "live") and confirm `useFestivalPhase()`-driven UI (banner/label) reflects it immediately.
- Clear the override and confirm the edition falls back to its derived phase.
- Run `pnpm test src/lib/festivalPhase.test.ts` — override-wins-over-every-derived-case and null-falls-through-to-derived cases pass.
- Insert two `set_ratings` rows for the same `(user_id, set_id)` as different users — both succeed; as the same user — second insert violates the unique constraint.
- Confirm no `supabase db push`/`db reset` was run; migration is timestamp-prefixed under `supabase/migrations/`.

Closes #139.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FUwNzs67viT4UnYJwqjSWo)_

### DIFF
--- a/src/pages/admin/festivals/PhaseOverrideControl.tsx
+++ b/src/pages/admin/festivals/PhaseOverrideControl.tsx
@@ -1,5 +1,3 @@
-import { X } from "lucide-react";
-import { Button } from "@/components/ui/button";
 import {
   Select,
   SelectContent,
@@ -14,6 +12,8 @@ type Props = {
   editionId: string;
   override: FestivalPhase | null;
 };
+
+const AUTOMATIC = "automatic";
 
 const PHASE_LABEL: Record<FestivalPhase, string> = {
   "pre-schedule": "Pre-Schedule",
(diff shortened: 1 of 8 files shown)


src/pages/admin/festivals/PhaseOverrideControl.tsx
```diff
@@ -0,0 +1,74 @@
...
+    <div className="flex flex-wrap items-center gap-2">
+      <Select
+        value={override ?? undefined}
+        disabled={isPending}
+        onValueChange={(value) => {
+          if (isFestivalPhase(value)) setOverride(value);
+        }}
+      >
+        <SelectTrigger className="w-[180px]">
+          <SelectValue placeholder="Phase: automatic" />
```
chiptus: Clearing doesn't revert this. Why not instead of a clear button, add a first option "Automatic"?

PR #147, opened 2026-07-10
feat(phases): add live now/next to schedule

chiptus:
Adds a Live-phase experience: a dedicated "Now" schedule view (one row per stage: current set + that stage's next set) that Live editions land on by default, backed by a pure injected-clock classifier, plus Live copy in the shared phase banner. Hidden and route-guarded outside Live or below reveal level `full`, so hidden times never leak; no background timer. Closes #138.

## Verification
- With an edition at reveal `full` and dates spanning today, open the edition: it lands on Schedule → **Now**, shows the Live banner, and lists one row per stage with the current set (pulsing dot, "until HH:mm") and that stage's next set ("at HH:mm"), in festival time.
- A stage whose next set starts later than another stage's still shows its *own* next set, not the globally nearest one.
- Click a set name in a row: navigates to the set detail (voting available there).
- Timeline and List views are unchanged and one tap away; the Now item only appears in the switcher when Live at full reveal.
- Navigate to `/schedule/now` on a non-Live edition or below reveal `full`: redirected to the timeline; no set times visible anywhere below `full`.
- Visit when nothing is playing and nothing is upcoming: the Now view shows a friendly empty state pointing at the timeline.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016AXf1EyBWLa985JwjYwvKX


src/lib/nowNext.ts
```diff
@@ -0,0 +1,61 @@
...
+export type NowNextClassification<T> = {
+  nowPlaying: T[];
+  next: T[];
+  laterPast: T[];
+};
+
+// Pure now/next classifier for Live mode. Compares UTC instants, so it is
+// festival-timezone-correct by construction; `now` is injected like in
+// getFestivalPhase. Sets missing either time (masked below reveal level
+// `full`, or unparseable) never classify — they are silently excluded.
```
chiptus: this comment says "how" the function is working.

replace with a jsdoc saying "what" the function is doing, what are we using it for?

src/pages/EditionView/tabs/ScheduleTab/now/NowBoard.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { stagesByEditionQuery } from "@/api/stages/useStagesByEdition";
+import { sortStagesByOrder } from "@/lib/stageUtils";
+import { NowStageRow } from "./NowStageRow";
+
+export function ScheduleTabNow() {
+  const { festival } = useFestivalEdition();
+  const { edition } = useRouteContext({
+    from: "/festivals/$festivalSlug/editions/$editionSlug/schedule/now",
+  });
+  const { data: sets = [], isLoading } = useSetsByEditionQuery(edition.id);
```
chiptus: why dpn't we use suspense?

src/routes/festivals/$festivalSlug/editions/$editionSlug/schedule/now.tsx
```diff
@@ -0,0 +1,30 @@
+import { createFileRoute, redirect } from "@tanstack/react-router";
+import { ScheduleTabNow } from "@/pages/EditionView/tabs/ScheduleTab/now/NowBoard";
+import { getFestivalPhase } from "@/lib/festivalPhase";
+import { canShowTime } from "@/lib/scheduleReveal";
+
+export const Route = createFileRoute(
+  "/festivals/$festivalSlug/editions/$editionSlug/schedule/now",
+)({
+  component: ScheduleTabNow,
```
chiptus, line 27: colocate component in this file

src/pages/EditionView/tabs/ScheduleTab/now/NowBoard.tsx
chiptus: Historical reason only: the board started as a section embedded above the Timeline, where suspending would have blocked the whole Schedule tab. Now that it's a dedicated route that concern is gone — switched both queries to `useSuspenseQuery` (with a sets prefetch in the route loader) in 8b91156, which also drops the manual loading branch.

---

src/routes/festivals/$festivalSlug/editions/$editionSlug/schedule/now.tsx
```diff
@@ -0,0 +1,149 @@
...
+  },
+});
+
+function ScheduleTabNow() {
+  const { festival } = useFestivalEdition();
+  const { edition } = Route.useRouteContext();
+  const { data: sets } = useSuspenseQuery(setsByEditionQuery(edition.id));
+  const { data: stages } = useSuspenseQuery(stagesByEditionQuery(edition.id));
+
+  const byStage = classifyNowNextByStage(sets, new Date());
```
chiptus: this will cause the component to rerender every second, is that what we want?

chiptus: It actually didn't rerender at all — `new Date()` in the render body schedules nothing, so the board was a snapshot from the last render (only refreshing on tab focus via query refetch). That's the opposite problem: it could go stale mid-set. Fixed in ab37718 by using the `useNow` hook (main already had one) so the classification tracks the clock once a minute while mounted — still no per-second churn.

---

PR #198, opened 2026-07-14
refactor(schedule): extract shared pure schedule filter

chiptus:
Extracts the duplicated day/time-of-day/stage filter predicates from Timeline and ListSchedule into one pure `filterScheduleDays` function in `src/lib/scheduleFilter.ts`, so upcoming vote-type filtering (parent PRD #188) can be added in a single place.
No behavior change; also removes a stray debug `console.log` from ListSchedule.

Closes #190

## Verification
- Unit tests: `pnpm vitest run src/lib/scheduleFilter.test.ts` covers day, festival-timezone time-of-day, and stage predicates plus combinations, without rendering the DOM.
- `pnpm run lint`, `pnpm run typecheck`, and `pnpm vitest run` (full suite, 364 tests) all pass.
- Manually: open the Schedule tab in both Timeline and List views, apply day/time/stage filters via the URL controls, and confirm the same sets are visible in both views as before this change.
- e2e runs in CI (no local Supabase available in this environment; no new e2e spec needed for a no-behavior-change prefactor).

---
_Generated by [Claude Code](https://claude.ai/code/session_01AziTqr3f12fxALYD6jhrW8)_

### DIFF
--- a/src/lib/scheduleFilter.test.ts
+++ b/src/lib/scheduleFilter.test.ts
@@ -4,43 +4,6 @@ import type { ScheduleDay, ScheduleSet } from "@/hooks/useScheduleData";
 
 const TIMEZONE = "Europe/Lisbon";
 
-function makeSet(overrides: Partial<ScheduleSet> = {}): ScheduleSet {
-  return {
-    id: "set-1",
-    name: "A set",
-    artists: [],
-    startTime: new Date("2024-07-15T10:00:00Z"),
-    ...overrides,
-  };
-}
-
-function makeDay(overrides: Partial<ScheduleDay> = {}): ScheduleDay {
-  return {
-    date: "2024-07-15",
-    displayDate: "Monday, Jul 15",
-    stages: [
-      {
-        id: "stage-1",
-        name: "Main Stage",
-        stage_order: 1,
-        sets: [makeSet()],
-      },
-    ],
-    ...overrides,
-  };
-}
-
-function baseCriteria(
-  overrides: Partial<ScheduleFilterCriteria> = {},
-): ScheduleFilterCriteria {
-  return {
-    day: "all",
-    time: "all",
-    stages: [],
-    ...overrides,
-  };
-}
-
 describe("filterScheduleDays", () => {
   describe("day predicate", () => {
     it("keeps all days when day is 'all'", () => {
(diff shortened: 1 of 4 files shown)


src/lib/scheduleFilter.test.ts
```diff
@@ -0,0 +1,298 @@
+import { describe, expect, it } from "vitest";
+import { filterScheduleDays, type ScheduleFilterCriteria } from "./scheduleFilter";
+import type { ScheduleDay, ScheduleSet } from "@/hooks/useScheduleData";
+
+const TIMEZONE = "Europe/Lisbon";
+
+function makeSet(overrides: Partial<ScheduleSet> = {}): ScheduleSet {
```
chiptus: Move helper functions to the bottom

src/lib/scheduleFilter.test.ts
```diff
@@ -0,0 +1,298 @@
...
+  describe("day predicate", () => {
+    it("keeps all days when day is 'all'", () => {
+      const days = [
+        makeDay({ date: "2024-07-15" }),
+        makeDay({ date: "2024-07-16" }),
+      ];
+
+      const result = filterScheduleDays(days, baseCriteria(), TIMEZONE);
+
+      expect(result[0].stages).toHaveLength(1);
```
chiptus, line 21: Should we test the dates?

src/lib/scheduleFilter.ts
```diff
@@ -0,0 +1,73 @@
...
+      return hour >= 18 && hour < 24;
+    default:
+      return true;
+  }
+}
+
+// Applies the day / time-of-day / stage predicates shared by both Schedule
+// views (Timeline and List). A day that doesn't match `criteria.day` is kept
+// as an entry with empty stages (rather than dropped), matching the
+// Timeline's day-strip layout, which needs every day present even when empty.
```
chiptus: That means that if we do drop them, the timeline will show only that day?

chiptus: Almost — the timeline already shows only that day either way. The strip's axis bounds come from the remaining sets' start/end times (`calculateTimeBoundaries` scans sets, not day entries), so an empty day entry contributes nothing and dropping it would render identically. The empty-entry shape was kept only to mirror the original inline logic, since this PR's contract was "no behavior change". I've reworded the comment in 50f12ee to say that (the old wording wrongly claimed the layout needs every day present). Happy to switch to dropping in a follow-up if you prefer the cleaner shape — both views handle it.

---

PR #201, opened 2026-07-14
feat(timeline): scroll position lives in url

chiptus:
Stacked on the #191 branch (`claude/188-191-timeline-geometry`) / PR. Adds a `scrollTo` search param that captures the moment centered in the Timeline viewport, with a pure precedence function deciding the mount position and a dedicated hook owning one-way scroll<->URL sync.

Closes #192

## Verification
- Load the timeline with no `scrollTo` in the URL; scroll the horizontal timeline and confirm `scrollTo` appears in the URL (~300ms after scrolling stops), rounded to 5 minutes, via history replace (back doesn't land on an intermediate scroll state).
- Load a timeline URL with `scrollTo` set; confirm the viewport centers on that moment on mount.
- Change the day filter with no `scrollTo` present; confirm the viewport centers on that day's start.
- Scroll the timeline and confirm the day/time/stage filters and rendered schedule do not recompute (no flicker/reflow) purely from the `scrollTo` write.
- Run `pnpm vitest run` — includes new unit tests for `resolveTimelineMountMoment` and `roundToNearestMinutes` in `src/lib/timelineMountMoment.test.ts`.
- New Playwright spec `tests/e2e/timeline-scroll.spec.ts` covers scroll-to-URL writing, mount centering, and back/reload restoring the viewport (runs in CI, not run locally in this environment).

---
_Generated by [Claude Code](https://claude.ai/code/session_01AziTqr3f12fxALYD6jhrW8)_

### DIFF
--- a/src/hooks/useTimelineScrollSync.ts
+++ b/src/hooks/useTimelineScrollSync.ts
@@ -87,6 +87,12 @@ export function useTimelineScrollSync({
 
     let debounceTimer: ReturnType<typeof setTimeout> | undefined;
 
+    container.addEventListener("scroll", handleScroll, { passive: true });
+    return () => {
+      container.removeEventListener("scroll", handleScroll);
+      if (debounceTimer) clearTimeout(debounceTimer);
+    };
+
     function handleScroll() {
       const programmaticLeft = programmaticScrollLeftRef.current;
       if (programmaticLeft !== null) {
@@ -116,11 +122,5 @@ export function useTimelineScrollSync({
         });
       }, SCROLL_DEBOUNCE_MS);
     }
-
-    container.addEventListener("scroll", handleScroll, { passive: true });
-    return () => {
-      container.removeEventListener("scroll", handleScroll);
-      if (debounceTimer) clearTimeout(debounceTimer);
-    };
   }, [scrollContainerRef, festivalStart, navigate]);
 }
--- a/src/hooks/useTimelineScrollSync.ts
+++ b/src/hooks/useTimelineScrollSync.ts
@@ -47,7 +47,10 @@ export function useTimelineScrollSync({
   const navigate = useNavigate({ from: route });
 
   const hasCenteredOnMountRef = useRef(false);
-  const suppressNextScrollEventRef = useRef(false);
+  // Position of the last programmatic scroll; scroll events reporting this
+  // position are ignored (a browser may fire more than one for a single
+  // scrollLeft write), so only genuine user scrolling reaches the URL.
+  const programmaticScrollLeftRef = useRef<number | null>(null);
 
   useLayoutEffect(() => {
     if (hasCenteredOnMountRef.current) return;
@@ -68,7 +71,7 @@ export function useTimelineScrollSync({
     );
 
     if (targetScrollLeft !== container.scrollLeft) {
-      suppressNextScrollEventRef.current = true;
+      programmaticScrollLeftRef.current = targetScrollLeft;
       container.scrollLeft = targetScrollLeft;
     }
     // Mount-only positioning: intentionally does not re-run when scrollTo/day
@@ -83,9 +86,13 @@ export function useTimelineScrollSync({
     let debounceTimer: ReturnType<typeof setTimeout> | undefined;
 
     function handleScroll() {
-      if (suppressNextScrollEventRef.current) {
-        suppressNextScrollEventRef.current = false;
-        return;
+      const programmaticLeft = programmaticScrollLeftRef.current;
+      if (programmaticLeft !== null) {
+        const el = scrollContainerRef.current;
+        if (el && Math.abs(el.scrollLeft - programmaticLeft) <= 1) {
+          return;
+        }
+        programmaticScrollLeftRef.current = null;
       }
 
       if (debounceTimer) clearTimeout(debounceTimer);
--- a/src/hooks/useTimelineScrollSync.ts
+++ b/src/hooks/useTimelineScrollSync.ts
@@ -42,7 +42,6 @@ export function useTimelineScrollSync({
   const { scrollTo, day } = useSearch({
     from: route,
     select: (search) => ({ scrollTo: search.scrollTo, day: search.day }),
-    structuralSharing: true,
   });
   const navigate = useNavigate({ from: route });
 
--- a/src/hooks/useTimelineScrollSync.ts
+++ b/src/hooks/useTimelineScrollSync.ts
@@ -1,0 +1,117 @@
+import { useEffect, useLayoutEffect, useRef } from "react";
+import { useNavigate, useSearch } from "@tanstack/react-router";
+import type { RefObject } from "react";
+import { offsetToTime, timeToOffset } from "@/lib/timelineCalculator";
+import {
+  resolveTimelineMountMoment,
+  roundToNearestMinutes,
+} from "@/lib/timelineMountMoment";
+
+const SCROLL_DEBOUNCE_MS = 300;
+const SCROLL_ROUND_MINUTES = 5;
+
+interface UseTimelineScrollSyncOptions {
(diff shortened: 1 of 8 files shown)


src/hooks/useTimelineUrlState.ts
```diff
@@ -8,8 +8,18 @@ export type TimeFilter = TimelineSearch["time"];
...
+  // identity and trigger consumers to recompute the filtered schedule.
   const state = useSearch({
     from: route,
+    select: (search) => ({
+      view: search.view,
+      day: search.day,
+      time: search.time,
+      stages: search.stages,
+    }),
+    structuralSharing: true,
```
chiptus: Check: what is structuralSharing

src/lib/searchSchemas.ts
```diff
@@ -40,6 +40,9 @@ export const timelineSearchSchema = z.object({
   day: z.string().catch("all"),
   time: z.enum(["all", "morning", "afternoon", "evening"]).catch("all"),
   stages: z.array(z.string()).catch([]),
+  // The moment (ISO datetime) centered in the timeline viewport. Absent by
```
chiptus: Do we really need this comment?

tests/e2e/timeline-scroll.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+    const historyLengthBeforeScroll = await page.evaluate(
+      () => window.history.length,
+    );
+
+    await scrollContainer.evaluate((el) => {
+      el.scrollLeft = el.scrollLeft + 400;
+    });
+
+    // No write yet: still inside the debounce window.
+    await page.waitForTimeout(100);
```
chiptus: Use fake clock, don't use wait for timeout

src/hooks/useTimelineUrlState.ts
chiptus: `structuralSharing` is a TanStack Router option on `useSearch`'s `select`: when the selected result is deep-equal to the previous one, it reuses the previous object/array references (via `replaceEqualDeep`) instead of returning fresh ones. It matters here because `select` builds a new object (with a new `stages` array) on every URL change — without structural sharing, a `scrollTo`-only write would hand the Timeline a new `stages` reference, re-triggering the filtering `useMemo` on every debounced scroll write. With it, `stages` keeps its identity as long as its contents are unchanged, so scroll writes never recompute the filtered schedule (the "scroll writes must not re-run filtering" requirement from #192).

---

src/hooks/useTimelineScrollSync.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    const targetScrollLeft = Math.max(
+      0,
+      timeToOffset(moment, festivalStart) - container.clientWidth / 2,
+    );
+
+    if (targetScrollLeft !== container.scrollLeft) {
+      programmaticScrollLeftRef.current = targetScrollLeft;
+      container.scrollLeft = targetScrollLeft;
+    }
```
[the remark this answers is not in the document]
chiptus: Partially taken. The clamp point was real: the ref now records the read-back `scrollLeft` after assignment (5fcab7c), so a browser-clamped mount scroll is still recognized as programmatic. The 20px point isn't: the hour-marker axis renders at the raw geometry offset — only set blocks and date labels carry a pre-existing `+20` — so the conversion here is consistent with the canonical axis mapping, and the round-trip (write moment → restore scrollLeft) is exact regardless. Aligning set blocks with the axis is a separate visual cleanup, out of scope for this PR.

---

src/hooks/useTimelineScrollSync.ts
```diff
@@ -0,0 +1,124 @@
...
+      }
+
+      if (debounceTimer) clearTimeout(debounceTimer);
+      debounceTimer = setTimeout(() => {
+        const el = scrollContainerRef.current;
+        if (!el) return;
+
+        const centerOffset = el.scrollLeft + el.clientWidth / 2;
+        const centerMoment = offsetToTime(centerOffset, festivalStart);
+        const rounded = roundToNearestMinutes(
```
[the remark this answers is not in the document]
chiptus: Not taken — same reasoning as the thread above: the hour-marker axis (the canonical time→pixel mapping) has no 20px offset; only set blocks/date labels carry a pre-existing `+20`. The conversion matches the axis, and the scrollTo round-trip restores the exact scroll position either way.

---

src/hooks/useTimelineScrollSync.ts
```diff
@@ -0,0 +1,124 @@
...
+    if (targetScrollLeft !== container.scrollLeft) {
+      programmaticScrollLeftRef.current = targetScrollLeft;
+      container.scrollLeft = targetScrollLeft;
+    }
+    // Mount-only positioning: intentionally does not re-run when scrollTo/day
+    // change afterwards (one-way ownership, URL -> scroll only on mount).
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [scrollContainerRef]);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
chiptus: Intentional, per the spec (issue #192 / PRD #188): mount precedence applies at mount only, and "filtering narrows, it never scrolls" — changing the day filter must never move the viewport. The day-filter rule exists only to pick the initial position when a page loads with `?day=` and no `scrollTo`.

---

src/hooks/useTimelineScrollSync.ts
```diff
@@ -0,0 +1,126 @@
...
+    const container = scrollContainerRef.current;
+    if (!container) return;
+
+    let debounceTimer: ReturnType<typeof setTimeout> | undefined;
+
+    function handleScroll() {
```
chiptus, line 95: move to the bottom of useEffect

src/hooks/useTimelineScrollSync.ts
```diff
@@ -0,0 +1,126 @@
...
+      const programmaticLeft = programmaticScrollLeftRef.current;
+      if (programmaticLeft !== null) {
+        const el = scrollContainerRef.current;
+        if (el && Math.abs(el.scrollLeft - programmaticLeft) <= 1) {
+          return;
+        }
+        programmaticScrollLeftRef.current = null;
+      }
+
+      if (debounceTimer) clearTimeout(debounceTimer);
```
chiptus, line 105: should we separate the debounce logic into its own hook?

chiptus: I'd hold off. This is the only debounce in the app (the #203 mini-map deliberately funnels through this same scroll listener rather than debouncing itself), and the timer here isn't a plain debounce — it doubles as "programmatic scroll animation still settling" suppression (mid-animation events keep resetting it, which is what makes `jumpTo` write exactly one settled value), and its cleanup is tied to the listener's unsubscribe. A generic `useDebouncedCallback` would extract ~4 lines of `clearTimeout`/`setTimeout` while splitting that lifecycle across two hooks. If a second debounce consumer shows up, extracting then is cheap — happy to do it now if you'd still prefer it.

---

src/hooks/useTimelineUrlState.ts
chiptus: should we just set structuralSharing to true as a general setting?

src/hooks/useTimelineUrlState.ts
```diff
@@ -7,8 +7,17 @@ export type TimeFilter = TimelineSearch["time"];
 export function useTimelineUrlState(tab: "timeline" | "list" = "timeline") {
   const route =
     `/festivals/$festivalSlug/editions/$editionSlug/schedule/${tab}` as const;
+  // Select only the filter params this hook exposes, with structural sharing,
+  // so a `scrollTo` write (from scroll syncing) doesn't change this object's
+  // identity and trigger consumers to recompute the filtered schedule.
```
chiptus: remove comment

PR #260, opened 2026-07-28
fix(components): replace confirm() with AlertDialog

chiptus:
Adds a shared ConfirmDialog and a `confirm(options): Promise<boolean>` helper (singleton store + `<ConfirmDialogHost />` mounted at the app root, modeled on the existing `useToast`/`<Toaster />` pattern), and swaps all 8 destructive-action `window.confirm()`/`confirm()` sites for it.
Cancel/dismiss resolves `false` and leaves data unchanged; confirming resolves `true` and the caller runs its mutation via `mutation.mutate(...)`, using the mutation's own `isPending` for loading state on the trigger button.

Closes #160

## Verification
- Delete a set note: dialog appears with note-specific copy; Cancel leaves it; Delete removes it.
- Delete a group invite in InviteManagement: dialog appears; confirm removes the invite.
- Remove a member from a group as the group creator: dialog appears; confirm removes the member, and only that member's button shows a disabled/loading state while removal is in flight.
- Leave a group as a non-creator member: dialog appears; confirm leaves the group.
- Delete a festival, stage, edition, and set in admin: each dialog shows action-specific copy; confirm removes the item.
- Confirm no `confirm(`/`window.confirm(` occurrences remain under `src/` (grep clean).

---
_Generated by [Claude Code](https://claude.ai/code/session_016zPmjZr8QrriCzLUgdGUt9)_

### DIFF
--- a/src/hooks/use-confirm.ts
+++ b/src/hooks/use-confirm.ts
@@ -1,0 +1,53 @@
+import { useEffect, useState } from "react";
+
+interface ConfirmOptions {
+  title: string;
+  description: string;
+  confirmLabel: string;
+}
+
+interface ConfirmRequest extends ConfirmOptions {
+  resolve: (confirmed: boolean) => void;
+}
+
+interface State {
+  request: ConfirmRequest | null;
+}
+
+let memoryState: State = { request: null };
+const listeners: Array<(state: State) => void> = [];
+
+function dispatch(state: State) {
+  memoryState = state;
+  listeners.forEach((listener) => listener(memoryState));
+}
+
+function confirm(options: ConfirmOptions) {
+  return new Promise<boolean>((resolve) => {
+    memoryState.request?.resolve(false);
+    dispatch({ request: { ...options, resolve } });
+  });
+}
+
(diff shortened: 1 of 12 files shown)


src/hooks/use-confirm.ts
```diff
@@ -0,0 +1,52 @@
...
+
+function respondConfirm(confirmed: boolean) {
+  memoryState.request?.resolve(confirmed);
+  dispatch({ request: null });
+}
+
+function useConfirm() {
+  const [state, setState] = React.useState<State>(memoryState);
+
+  React.useEffect(() => {
```
chiptus: import useEffect

chiptus: Done — switched to `import { useEffect, useState } from "react"` in 8ffc3c6. Also fixed a real bug while in there: `confirm()` was overwriting `memoryState.request` without resolving the previous one first, so if two confirms fired before the first was answered, the first caller's promise would hang forever. Now it resolves any pending request as `false` before dispatching the new one.

---

Issue #157, opened 2026-07-13
refactor: finish mutateAsync → mutation.mutate() migration and add error toasts for silent save/create failures


chiptus:
## Problem
CLAUDE.md's convention is `mutation.mutate(variables, {onSuccess, onError})`, not `try { await x.mutateAsync(...) } catch {}`. 10 files still use the old pattern, and `MUTATE_ASYNC_CONVERSION_PLAN.md` is stale — it references files (`useOfflineVoting.ts`, `useOfflineNotes.ts`) that no longer exist in the repo, while omitting files that still need conversion.

Easiest targets (single mutation, no branching):
- `src/pages/admin/festivals/StageManagement/CreateStageDialog.tsx:24`
- `src/pages/admin/festivals/StageManagement/EditStageDialog.tsx:29`
- `src/pages/groups/GroupDetail.tsx:201`
- `src/components/invite/useInviteValidation.ts:56`

(6 more files also need conversion — see full audit in this repo's chat history / re-run a grep for `mutateAsync` under `src/`.)

## Fix
Convert each to `mutation.mutate(variables, { onSuccess, onError })`. Bonus: none of these currently have `onError` handling, so converting also fixes silent-failure UX for stage create/edit and member removal.

## Related
Update or delete the stale `MUTATE_ASYNC_CONVERSION_PLAN.md` as part of this (or see separate housekeeping ticket).

chiptus:
Claiming this for the autonomic fix worker.

- Timestamp: 2026-09-12T00:00:00Z (session start)
- Branch: `fix-157/finish-mutate-migration`

Note: the specific "easiest targets" listed in the issue have already been converted and `MUTATE_ASYNC_CONVERSION_PLAN.md` no longer exists, but 7 `mutateAsync` call sites remain across 5 files (`AddArtistDialog.tsx`, `FestivalDialog.tsx` ×2, `FestivalLogoDialog.tsx` ×2, `FestivalEditionManagement.tsx` ×2) — converting those.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YNxZeiFoAMqPB9C4QLQzLw)_

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #267, opened 2026-07-30
refactor(mutations): finish mutateAsync migration

chiptus:
Converts the remaining try/mutateAsync call sites in `CreateStageDialog` and `EditStageDialog` to `mutation.mutate(variables, {onSuccess, onError})` per CLAUDE.md convention, closing #157.
`StageForm`'s submitting state now comes straight from the mutation's `isPending` instead of a local `useState` the caller had to manage.

## Verification
- Create a stage in Stage Management; button shows a pending state while saving, then the dialog closes on success.
- Edit a stage; button shows a pending state while saving, then the dialog closes on success.
- Simulate a stage mutation failure (e.g. offline); error toast appears instead of a silent/unhandled rejection.

---
_Generated by [Claude Code](https://claude.ai/code/session_016zPmjZr8QrriCzLUgdGUt9)_

### DIFF
--- a/src/components/invite/useInviteValidation.ts
+++ b/src/components/invite/useInviteValidation.ts
@@ -49,7 +49,7 @@ export function useInviteValidation(inviteToken: string | undefined) {
     }
   }, [inviteValidation, toast]);
 
-  function acceptInvite(userId: string): Promise<boolean> {
+  function useInvite(userId: string): Promise<boolean> {
     if (!inviteToken) return Promise.resolve(false);
 
     return new Promise((resolve) => {
@@ -67,7 +67,7 @@ export function useInviteValidation(inviteToken: string | undefined) {
             resolve(true);
           },
           onError: (error) => {
-            console.error("failed accepting invite", error);
+            console.error("failed validating invite", error);
             resolve(false);
           },
         },
@@ -80,7 +80,7 @@ export function useInviteValidation(inviteToken: string | undefined) {
     inviteValidation,
     isValidating,
     validationError: validationError?.message || null,
-    acceptInvite,
+    useInvite,
     hasValidInvite: inviteValidation?.is_valid === true,
   };
 }
--- a/src/components/invite/useInviteValidation.ts
+++ b/src/components/invite/useInviteValidation.ts
@@ -49,7 +49,7 @@ export function useInviteValidation(inviteToken: string | undefined) {
     }
   }, [inviteValidation, toast]);
 
-  function useInvite(userId: string): Promise<boolean> {
+  function acceptInvite(userId: string): Promise<boolean> {
     if (!inviteToken) return Promise.resolve(false);
 
     return new Promise((resolve) => {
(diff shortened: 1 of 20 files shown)


src/components/invite/useInviteValidation.ts
```diff
@@ -49,25 +49,30 @@ export function useInviteValidation(inviteToken: string | undefined) {
     }
   }, [inviteValidation, toast]);
 
-  async function useInvite(userId: string) {
-    if (!inviteToken) return false;
+  function useInvite(userId: string): Promise<boolean> {
```
chiptus, line 52: This is a function within a hook? Does it need to be a hook?

src/pages/admin/festivals/StageManagement/CreateStageDialog.tsx
```diff
@@ -20,12 +20,19 @@ export function CreateStageDialog({ editionId }: CreateStageDialogProps) {
   const [isOpen, setIsOpen] = useState(false);
   const createStageMutation = useCreateStageMutation();
 
-  async function handleSubmit(data: StageFormData) {
-    await createStageMutation.mutateAsync({
-      ...data,
-      festival_edition_id: editionId,
+  function handleSubmit(data: StageFormData) {
```
chiptus, line 23: Why do we need a promise?

src/components/invite/useInviteValidation.ts
chiptus: It's actually about `StageForm`'s submit-spinner contract, not the mutation naming — its `onSubmit` prop is typed `(data) => Promise<void>` and it `await`s that call to know when to flip `isSubmitting` off (see `StageForm.tsx`'s `handleFormSubmit`). Since `mutate()` itself is fire-and-forget, `handleSubmit` wraps it in a `Promise` that resolves on `onSettled` so the form can await it and clear its own loading state. If you'd rather `StageForm` not require a `Promise`-returning `onSubmit`, that'd need a small change to how it tracks its own submitting state — happy to do that as a follow-up if preferred, but wanted to keep this PR scoped to the `mutateAsync` → `mutate` conversion.

---

src/pages/admin/festivals/StageManagement/CreateStageDialog.tsx
chiptus: `StageForm`'s `onSubmit` prop is typed `(data) => Promise<void>` — it `await`s that call in `handleFormSubmit` to know when to flip its own `isSubmitting` spinner off. Since `mutation.mutate()` is fire-and-forget (no returned promise), `handleSubmit` wraps it in a `new Promise` that resolves in `onSettled`, so the form can still `await` it. Kept this PR scoped to the `mutateAsync` → `mutate` conversion rather than also changing how `StageForm` tracks submitting state — happy to follow up on that separately if you'd rather avoid the wrapper.

---

chiptus: Done — `StageForm` now takes an `isSubmitting` prop instead of awaiting a `Promise`-returning `onSubmit`, and `CreateStageDialog`/`EditStageDialog` pass `mutation.isPending` directly. `handleSubmit` is back to a plain `mutate(...)` call, no wrapper.

---

PR #273, opened 2026-07-31
feat(groups): add active group switcher

### LINKED ISSUE #124: Group switcher + Vote Perspective on Artists tab
## Parent

chiptus/UpLine#122

## What to build

A single, persistently visible switcher (not per-tab) listing "Everyone" plus every Group the user belongs to. Selecting any entry immediately commits it as the new Active Group everywhere — there is no separate, non-persisting "preview" mode; every selection is a real change, saved via the foundation built in the prerequisite issue.

This switcher replaces the existing Artists-tab-only group dropdown. Wire the Artists tab's existing group-perspective feature onto it, renaming the concept **Vote Perspective**: a two-state toggle, Everyone ↔ Active Group, governing which votes feed a set's aggregate rating and popularity score (and therefore sort order). Vote Perspective never hides sets — only re-scores and re-sorts them. It defaults to Active Group when one exists, else Everyone. There is no "Me" option on this toggle.

Retire the per-tab `groupId` URL parameter — group identity now comes from the persisted Active Group, not the URL.

As part of touching this code (this is the natural, and only, place to do it — don't create a separate prefactor-only issue for it), extract the ad hoc group-membership vote filter that currently lives inline inside the Artists tab's set-filtering logic into one shared, pure, framework-free function: given a set's full list of votes, a scope selection, the Active Group's member IDs, and the current user's ID, it returns just the votes that count for that scope. This function is what the next issue (Vote Scope on the Schedule tab) will also call — build it general enough to serve both Vote Perspective's "Everyone vs Group" split and the next issue's "Me vs Group" split, without hard-coding either UI's specific default.

## Acceptance criteria

- [ ] A single switcher component is visible while browsing an edition (not scoped to one tab), listing Everyone + every Group the current user belongs to.
- [ ] Selecting any entry in the switcher persists it as the Active Group (calls the foundation from the prerequisite issue) and takes effect immediately, without a page reload.
- [ ] The old Artists-tab-only group dropdown component is removed; the `groupId` URL search param is removed from the relevant search schema and no longer read anywhere.
- [ ] Artists tab ratings, popularity scores, and sort order default to the Active Group's votes when one exists, and to Everyone when it doesn't.
- [ ] A visible, one-tap Vote Perspective toggle lets the user switch between Everyone and the Active Group on the Artists tab.
- [ ] The shared vote-resolution function is extracted as a pure function (no React, no Supabase client) with unit tests covering: Everyone scope, Me scope, Group scope with the current user as a member, Group scope with the current user not a member, a set with no votes, and a set with votes only from users outside the group. Tests follow the existing style in `voteConfig.test.ts` (Vitest, plain data in/out, no mocking).
- [ ] Admin/Core Team curation views (edition/artist editing) are unaffected by a curator's own Active Group setting.
- [ ] Switching Active Group never changes what other users can see of the current user's votes (no regression to vote visibility/read access).

## Blocked by

- chiptus/UpLine#123

chiptus:
Adds a persistent header switcher (Everyone + your groups) that sets the Active Group everywhere, and retires the Artists-tab-only group filter in favor of a two-state Vote Perspective toggle built on a shared, tested vote-scope resolver.
Selecting a group in the header re-scores and re-sorts the Artists tab immediately, without a reload.

Closes #124

## Verification
- Sign in as a user in 2+ groups; open the header switcher, pick a group, and confirm the button updates to that group's name immediately (no reload).
- On the Artists tab, with an Active Group set, confirm the Vote Perspective toggle defaults to the group and toggling to Everyone re-sorts/re-scores sets without hiding any.
- Sign in as a user in zero groups; confirm the header still shows the "Create/Join a Group" CTA and the Artists tab has no Vote Perspective toggle (always Everyone).
- `pnpm test` passes, including new `src/lib/voteScope.test.ts` covering Everyone/Me/Group scopes, a member vs. non-member group, no votes, and votes only from outside the group.
- `pnpm run typecheck` and `pnpm run lint` pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VGSj8prZCB1qAtR2Wz3qKL)_


src/components/layout/AppHeader/GroupsIndicator.tsx
```diff
@@ -74,7 +66,62 @@ function GroupsIndicatorContent({
...
+
+  function handleSelect(groupId: string | null) {
+    if (groupId === (activeGroupId ?? null)) {
+      return;
+    }
+    setActiveGroupMutation.mutate({ userId, groupId });
+  }
+
+  return (
+    <DropdownMenu>
```
chiptus: refactor the dropdown into a separate component

src/lib/voteScope.test.ts
```diff
@@ -0,0 +1,83 @@
...
+      resolveVotesForScope({
+        votes,
+        scope: "group",
+        groupMemberIds: ["user-1", "user-2"],
+        currentUserId: "user-1",
+      }),
+    ).toEqual([votes[0], votes[1]]);
+  });
+
+  it("excludes the current user's own vote from the group scope when they are not a member", () => {
```
chiptus, line 43: does that mean we allow a user to see votes of other groups?

src/lib/voteScope.test.ts
```diff
@@ -0,0 +1,83 @@
...
+      }),
+    ).toEqual([]);
+  });
+
+  it("accepts group member ids as a Set", () => {
+    expect(
+      resolveVotesForScope({
+        votes,
+        scope: "group",
+        groupMemberIds: new Set(["user-3"]),
```
chiptus: why do we need to support this use case?

src/lib/voteScope.test.ts
chiptus: No new read access here — this is purely a client-side re-scoring of votes the client already has. `set.votes` (every vote's `user_id`/`vote_type`) is fetched for every authenticated user regardless of scope; `resolveVotesForScope` just decides which of those already-fetched votes count toward the displayed rating/popularity, it never issues a new query or reveals anything not already in the response.

The function also deliberately doesn't check "is `currentUserId` in `groupMemberIds`" for the `group` branch — group scope means "these members' votes," independent of whether the viewer happens to be one of them. That's required because the next issue (Schedule tab's Me/Group split) reuses this same function, and its "Group" reading needs to work the same way. In this PR's actual usage (Artists tab Vote Perspective) `group` scope is only ever invoked with the viewer's own Active Group, since `activeGroupId` always comes from `useActiveGroup`, which only ever resolves to a group the current user belongs to — so the "not a member" path never fires from live UI code today. It's tested because issue #124's acceptance criteria explicitly listed it as one of the six required scenarios, to lock in the general contract for that future caller.

Vote read-access/visibility itself is unchanged — still governed by existing RLS, not by this function.

---

src/components/layout/AppHeader/ActiveGroupSwitcher.tsx
```diff
@@ -0,0 +1,189 @@
+import {
```
chiptus: Decompose this file into smaller files

chiptus: Split into three files in `b552980`: `scopeDisplay.ts` (pure label/icon/key helpers), `ScopeMenuBody.tsx` (the dropdown list), and the trimmed `ActiveGroupSwitcher.tsx` (now 91 lines — just the trigger + back-to-default pill).

---

src/components/layout/AppHeader/ActiveGroupSwitcher.tsx
```diff
@@ -0,0 +1,89 @@
...
+          >
+            <Link to="/groups">
+              <Settings className="h-4 w-4 mr-2" />
+              Manage groups
+            </Link>
+          </DropdownMenuItem>
+        </DropdownMenuContent>
+      </DropdownMenu>
+
+      {isOverridden && (
```
chiptus: I don't think we need this button, just adds jumping to the ui and the default should be the top of the menu so easy to reset

src/api/groups/useProfileFieldMutation.ts
```diff
@@ -0,0 +1,46 @@
...
+  column: K;
+  errorMessage: string;
+}) {
+  const queryClient = useQueryClient();
+  const { toast } = useToast();
+
+  return useMutation({
+    mutationFn: async (variables: {
+      userId: string;
+      value: ProfileUpdate[K];
```
chiptus: Why should the mutation update not nly one value?

src/components/layout/AppHeader/GroupSwitcher/ScopeMenuBody.tsx
```diff
@@ -0,0 +1,107 @@
+import { Star } from "lucide-react";
```
chiptus, line 1: This is better, but since it's only used in the group switcher it should be in a folder specifically for the group switcher or name prefix should be GroupSwitcher

src/components/layout/AppHeader/GroupSwitcher/ScopeMenuBody.tsx
```diff
@@ -0,0 +1,107 @@
...
+  current: PinnedScope;
+  onSelect: (scope: PinnedScope) => void;
+}) {
+  const pinnedGroupId = pinned.kind === "group" ? pinned.groupId : undefined;
+  const otherGroups = groups.filter((group) => group.id !== pinnedGroupId);
+  const otherScopeKinds = (["everyone", "me"] as const).filter(
+    (kind) => pinned.kind !== kind,
+  );
+
+  function isPinned(scope: PinnedScope) {
```
chiptus: Move fns to the bottom

src/contexts/ActiveScopeContext.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+function AuthedActiveScopeProvider({
+  userId,
+  children,
+}: {
+  userId: string;
+  children: ReactNode;
+}) {
+  const { profile } = useAuth();
+  const { data: groups = [], isLoading } = useQuery(userGroupsQuery(userId));
```
chiptus: Why don't we use suspense?

src/contexts/ActiveScopeContext.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+  const pinned = resolvePinnedScope({
+    activeGroupId: profile?.active_group_id,
+    activeScope: profile?.active_scope,
+    groupIds,
+  });
+
+  const current = override ?? pinned;
+
+  const setActiveGroupMutation = useProfileFieldMutation({
```
chiptus: One update profile mutation is enough

src/contexts/ActiveScopeContext.tsx
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  const value: ActiveScopeContextValue = {
+    isLoading,
+    groups,
+    hasGroups: groups.length > 0,
+    activeGroupId,
+    pinned,
+    current,
+    selectScope,
```
chiptus, line 99: Why don't we use one "selects cope" fn that has pinned boolean? Instead of two?

src/pages/groups/Groups/SignInRequired.tsx
```diff
@@ -8,13 +8,17 @@ import {
 import { Button } from "@/components/ui/button";
 import { Link } from "@tanstack/react-router";
 
-export function SignInRequired() {
+export function SignInRequired({
+  description = "Please sign in to manage groups",
```
chiptus: Hwy is this the default description?

src/pages/Settings/settingsToggleItemClassName.ts
```diff
@@ -0,0 +1,2 @@
+export const settingsToggleItemClassName =
```
chiptus: Why is this a string instead of a component?

supabase/migrations/20260814140000_add_active_group_selected_to_profiles.sql
```diff
@@ -0,0 +1,5 @@
+-- Distinguishes "never touched the switcher" from "explicitly selected
```
chiptus: Do we need this migration? As it's removed in the next?

src/pages/Settings/ActiveGroupSetting.tsx
```diff
@@ -0,0 +1,58 @@
...
+  userId,
+  groups,
+}: {
+  userId: string;
+  groups: Group[];
+}) {
+  const { activeGroupId, clearOverride } = useActiveScope();
+  const mutation = useProfileFieldMutation();
+
+  function setActiveGroup(groupId: string) {
```
chiptus: move function to bottom

PR #278, opened 2026-08-01
ci(e2e): split workflows, tag-based smoke tests

chiptus:
Splits e2e workflows into reusable steps (e2e-run.yml, e2e-report.yml) shared by smoke, nightly-full, and a new label-triggered full suite, and switches smoke test selection from hardcoded filenames to a `@smoke` Playwright tag.

## Verification
- Push to a PR without the `full-e2e` label; only the smoke job runs (17 chromium tests via `--grep @smoke`) and posts a PR summary comment.
- Add the `full-e2e` label to a PR; the full sharded suite runs and posts its own summary comment.
- Push a new commit to that same PR while the label is still attached; full suite re-runs.
- Remove the label and push again; full suite no longer runs.
- Trigger `E2E Full Suite (Nightly)` via workflow_dispatch; confirm it still shards, merges reports, and files a GitHub issue on failure.
- Tag a new test `@smoke` in any spec file and confirm it's picked up by smoke without touching workflow YAML.


.github/workflows/e2e-nightly.yml
```diff
@@ -0,0 +1,88 @@
...
+              });
+            } else {
+              await github.rest.issues.create({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                title,
+                body,
+                labels: [label],
+              });
+            }
```
chiptus: should we move this to a script file?

PR #279, opened 2026-08-01
refactor(sets): extract panel from route index

chiptus:
Splits the sets tab route file by moving group-filtering/panel logic into a dedicated `FilteredSetsPanel` component, leaving the route file with just route config and the top-level component.

## Verification
- Load the sets tab for an edition with no group filter; sets render as before.
- Apply a group filter for the first time; group members load and sets filter correctly.
- Lock sort order; order persists via URL state.

---
_Generated by [Claude Code](https://claude.ai/code/session_0129k1hrC1x4dZyVAbJ157j9)_


src/pages/EditionView/tabs/VoteTab/FilteredSetsPanel.tsx
```diff
@@ -0,0 +1,84 @@
+import { useMemo } from "react";
```
chiptus, line 1: why in ArtistTab and not SetsTab?

chiptus: Followed the existing convention — the folder is already named `ArtistsTab` and holds all the sets-related pieces (`SetsPanel`, `SetCard`, `useSetFiltering`, `filters/`, etc.), even though the route itself is `sets/index.tsx`. Didn't want to rename the directory as part of this refactor since that's a broader, unrelated change. Happy to rename `ArtistsTab` → `SetsTab` in a follow-up if you'd like that cleanup.

---

PR #289, opened 2026-08-04
fix(slugs): disambiguate duplicate set and artist slugs

chiptus:
Disambiguates a colliding set/artist slug with a numeric counter (`-2`, `-3`, ...) instead of inserting a duplicate. Generation and dedup now happen entirely at the DB layer — a `BEFORE INSERT` trigger derives the slug (via `public.slugify`) and resolves collisions under a transaction-scoped advisory lock, backed by a unique constraint (sets: per festival edition; artists: global, already existed) so concurrent creates can't slip through either. The frontend no longer computes slugs client-side.

Closes #135

## Verification
- Create a set named the same as an existing set in the same edition; new set gets slug `-2` and its detail page loads.
- Create a set with a name colliding with a set in a *different* edition; slug is unchanged (no suffix).
- Create an artist named the same as an existing artist; new artist gets slug `-2` and its detail page loads.
- Run the edge function integration tests (`supabase/functions/commit-schedule/commit-schedule.test.ts`) — `two sets with the same name get distinct slugs` exercises the DB-level dedup via the `commit_schedule` RPC.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LBa5F78BNbcTrjeAbYt5J6)_


src/api/artists/useCreateArtist.ts
```diff
@@ -27,12 +26,13 @@ async function createArtist(
   },
 ): Promise<Artist> {
   const { genre_ids, ...artist } = artistData;
-  // First, create the artist without slug
   const { data, error } = await supabase
     .from("artists")
     .insert({
       ...artist,
-      slug: generateSlug(artist.name),
+      // The artists_dedupe_slug trigger derives the real slug from `name`
```
chiptus: Replace 2 line comment with a simple 1 line "auto generated by server"

supabase/migrations/20260804103506_commit_schedule_slugify_delegates.sql
```diff
@@ -0,0 +1,11 @@
+-- commit_schedule__slugify duplicated the same regex logic public.slugify()
+-- now has (added in the previous migration). Delegate instead of
+-- maintaining two copies that could drift apart.
+CREATE OR REPLACE FUNCTION public.commit_schedule__slugify(p_name TEXT)
```
chiptus: Why don't we call the slugify fn directly?

src/api/artists/useCreateArtist.ts
chiptus: Done in `e2d8c6d` — shortened to `slug: "", // auto-generated by the artists_dedupe_slug DB trigger` (same for the sets hook).

---

supabase/migrations/20260804103506_commit_schedule_slugify_delegates.sql
chiptus: Fair — no good reason to keep it. Updated `commit_schedule__upsert_stages` and `commit_schedule__create_sets` to call `public.slugify()` directly and dropped `commit_schedule__slugify` entirely in `e2d8c6d`.

---

Issue #280, opened 2026-08-01
Integration-test tier scaffold + harness core, proven by useGenresQuery


chiptus:
## Parent

#64

## What to build

A new "integration test" tier, distinct from unit tests, that runs against a real local Supabase instance instead of mocking. This ticket establishes the tier and its core harness, proven end-to-end by migrating the `useGenresQuery` hook test off `vi.mock`/`vi.spyOn` onto it.

- A `*.integration.test.ts` file-naming convention for this tier.
- A separate Vitest config and `pnpm test:integration` script for it (analogous to how `test:e2e` is separate from `test`), excluded from the default `vitest.config.ts` so the existing fast per-PR unit-test run is unaffected.
- The shared harness core: a real Supabase client wired to the local test instance, plus a shared `afterEach` cleanup hook that later tickets' fixture factories register against. No fixture factories or auth helper yet — just the plumbing.
- `useGenresQuery`'s test migrated to run through this harness against real data, replacing its current `vi.mock`/`vi.spyOn` of `useSuspenseQuery`.
- `src/test/setup.ts`'s comment claiming tests mock the relevant query hooks updated to reflect the new two-tier split.

## Acceptance criteria

- [ ] `pnpm test:integration` exists and runs only `*.integration.test.ts` files, against a manually-started local Supabase instance.
- [ ] `pnpm test` (default) does not pick up `*.integration.test.ts` files and is otherwise unaffected.
- [ ] The shared harness module provides a real Supabase client and a shared cleanup hook, ready for other tickets' factories to register against.
- [ ] `useGenresQuery`'s test runs through the harness against real Supabase data (success and empty-result cases), with no mock of the hook or `useSuspenseQuery`.
- [ ] `src/test/setup.ts`'s stale mocking comment is corrected.

## Blocked by

- None — can start immediately.

PR #290, opened 2026-08-06
fix(invites): consume invite tokens on signup

### LINKED ISSUE #165: refactor: remove duplicate invite-token redemption logic from AuthContext
## Problem
`src/contexts/AuthContext.tsx:66-107` contains its own inline implementation of invite-token redemption (call `use_invite_token` RPC, check `result.success`/`result.message`, toast, clear URL param) inside the `onAuthStateChange` listener — a second, independent code path for the same operation `useInviteMutation()` already provides (`src/api/invite-validation/useInviteValidationQuery.ts:43-92`), with near-identical response-shape handling hand-copied between the two.

This embeds invite/group domain logic inside the auth seam, which should own session/profile only. If the RPC's response contract changes, both call sites need to be kept in sync by hand.

## Fix
`onAuthStateChange` calls `useInviteMutation()` instead of re-implementing the RPC call and response handling inline.

## Architecture note
Filed from an architecture review (module/interface/depth vocabulary, see `/codebase-design`). AuthContext's interface shrinks back to session/profile only once this is routed through the existing seam.

### LINKED ISSUE #268: Invite acceptance is never actually consumed — new/existing users never join the group
## Problem

Group invite links validate and display correctly, but accepting an invite never actually adds the user to the group. The `use_invite_token(token, user_id)` RPC — the only thing that inserts a `group_members` row — is never called anywhere in the app.

## Trace

1. Someone clicks a group invite link → lands at `/?invite=<token>` → `RootContent` (`src/routes/__root.tsx`) validates the token via `useInviteValidation` and renders `InviteLandingPage` ("You're Invited! Join Group").
2. They sign up through `AuthDialog` (magic link or OTP). `useSignInWithOtpMutation` (`src/api/auth/useSignInWithOtpMutation.ts`) stashes the invite token in `user_metadata.invite_token` and in the magic-link redirect URL (`?invite=<token>`), but that's all it does with it.
3. `InviteLandingPage`'s `onSignupSuccess` prop is wired to a no-op — `onSignupSuccess={() => {}}` in `src/routes/__root.tsx:108`.
4. `useInviteValidation.ts` exposes an `acceptInvite(userId)` function that wraps `useInviteMutation()` (which calls the `use_invite_token` RPC) — but **nothing in the codebase calls `acceptInvite`**. Confirmed via full-repo search: only `__root.tsx` calls `useInviteValidation()`, and it destructures just `{ inviteValidation, isValidating, hasValidInvite }`.
5. There's no database trigger that consumes `invite_token` from `auth.users` metadata either (checked `supabase/migrations/*create_functions.sql` — `use_invite_token` is designed to be called explicitly from the client, it's `SECURITY DEFINER` but not trigger-attached).

Net effect: a user can go through the entire "you're invited → sign up" flow and never end up in `group_members`. The invite is validated but never redeemed.

## Fix

Wire `onSignupSuccess` (and the equivalent path for an existing user who clicks an invite link while logged out then logs in) to actually call `acceptInvite(user.id)` once a session exists, and surface success/failure (toast, redirect into the group, etc.). Needs to handle both:
- A brand-new user completing OTP verification inside `AuthDialog`.
- The magic-link click landing back on `/?invite=<token>` with a fresh session (no callback fires here — this needs an effect keyed on `user` + `inviteToken` becoming available).

## Acceptance criteria

- [ ] A new user who signs up via a group invite link ends up as a `group_members` row for that group after completing signup (magic link and OTP paths both work).
- [ ] An existing (logged-out) user who clicks an invite link, then logs in, is added to the group.
- [ ] A user already in the group who reuses an invite link doesn't error or duplicate (already handled server-side by `use_invite_token`, just needs to surface cleanly).
- [ ] Some visible confirmation (toast and/or redirect) that they joined the group.
- [ ] Existing `InviteLandingPage` "Invalid Invite" / expired / deactivated states are unaffected.

## Context

Found while reviewing PR #267 (`mutateAsync` → `mutate` migration), which touched `useInviteValidation.ts`'s `acceptInvite` (formerly `useInvite`) without being aware it had no caller.

chiptus:
Wires the never-called `use_invite_token` RPC into a `useInviteAcceptance` effect keyed on user + invite token, so invite links actually add users to the group (magic-link, OTP, and already-signed-in paths). Also fixes `InviteLandingPage` sending `invite_id` instead of the real token into the auth flow, and removes the deadlock-prone RPC call from `AuthContext`'s `onAuthStateChange`.

Closes #268
Closes #165

## Verification
- Generate a group invite link, open it logged out, sign up via the 6-digit OTP code; you land in the app with a "Welcome to {group}" toast, the group appears in your groups, and `?invite=` is cleared from the URL.
- Same flow but click the magic link in the email instead of entering the code; joining works the same way.
- As an existing logged-out user, open an invite link and log in; you're added to the group.
- Re-open an invite link for a group you're already in; a friendly "Already a member" notice shows, no error and no duplicate membership.
- Open an expired/deactivated/unknown invite link; the existing "Invalid Invite" states render unchanged.

---
_Generated by [Claude Code](https://claude.ai/code/session_011AAZzVYMN7csUv8kehZEsp)_


src/api/invite-validation/useAcceptInviteMutation.ts
```diff
@@ -0,0 +1,56 @@
...
+}: AcceptInviteParams): Promise<AcceptInviteResult> {
+  const { data, error } = await supabase.rpc("use_invite_token", {
+    token,
+    user_id: userId,
+  });
+
+  if (error) {
+    console.error("Error using invite:", error);
+    throw new Error("Failed to join group");
+  }
```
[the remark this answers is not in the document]
chiptus: Keeping the generic message intentionally: this string ends up verbatim in a user-facing toast, and the raw Supabase/PostgREST error text (network failures, RLS or SQL details) isn't something an end user can act on. The underlying error is preserved via `console.error` right above for debugging. Business-outcome failures from the RPC itself (e.g. an expired or invalid invite) do surface their specific message — only transport/unexpected-shape errors fall back to the generic copy.

---

src/components/invite/useInviteAcceptance.test.tsx
```diff
@@ -0,0 +1,262 @@
+import { describe, expect, it, vi, beforeEach } from "vitest";
```
chiptus: Feels like we're checking mostly mocks instead of the actual behavior. Is this test valid?

src/contexts/AuthContext.tsx
```diff
@@ -42,86 +41,33 @@ export function AuthProvider({ children }: AuthProviderProps) {
   const [inviteToken, setInviteToken] = useState<string | undefined>();
   const [groupName, setGroupName] = useState<string | undefined>();
 
-  const { toast } = useToast();
   const profileQuery = useProfileQuery(user?.id);
   const profile = profileQuery.data;
 
   useEffect(() => {
-    // Set up auth state listener first
+    // Set up auth state listener first. Invite acceptance is handled by
```
chiptus: We don't need the extra comment

chiptus (commented):
Why do we need all the different invite hooks?

src/components/invite/useInviteAcceptance.test.tsx
chiptus: Fair challenge — the original version tested a hook in isolation, so half the assertions were about wiring. I've reworked it (now `useInviteFlow.test.tsx`): the test drives the whole flow through the real TanStack Query hooks (`useInviteValidationQuery` + `useAcceptInviteMutation`), and only the supabase client's `rpc` is mocked, dispatching by function name like the real backend would. What's under test is real behavior: validation gating (an expired invite toasts and never calls `use_invite_token`), waiting for a session before consuming, once-per-token dedup, already-member handled without an error, retry after failure, and URL cleanup. Router/toast mocks are just observation points for outcomes.

The remaining mock is the network edge itself — replacing that means running against local Supabase, which is exactly what #280's integration-test tier introduces. This test is a natural candidate to migrate onto that harness once it lands.

---

src/components/invite/useInviteFlow.ts
```diff
@@ -0,0 +1,98 @@
...
+  user: User | null,
+) {
+  const { toast } = useToast();
+  const navigate = useNavigate();
+
+  const {
+    data: inviteValidation,
+    isLoading: isValidating,
+    error: validationError,
+  } = useInviteValidationQuery(inviteToken || null);
```
chiptus: Instead of destructuring, we can call the value inviteQuery and then it's clear and less variables

src/components/invite/useInviteFlow.ts
```diff
@@ -0,0 +1,92 @@
...
+    if (inviteQuery.error) {
+      toast({
+        title: "Invalid Invite",
+        description: "This invite link is not valid",
+        variant: "destructive",
+      });
+    }
+  }, [inviteQuery.error, toast]);
+
+  useEffect(() => {
```
chiptus: Move switch code to a function getError. And aren't the two use effects in danger of calling toast twice?

chiptus: Both done in 20c871a. Good catch on the double toast — with `staleTime`/`gcTime` at 0, a refetch failure can leave a stale `data` alongside a fresh `error`, so the two effects could each fire. They're now merged into one effect driven by a single `getError(error, validation)` helper (error wins, then the reason switch), so at most one toast per state change.

---

PR #291, opened 2026-08-14
fix(supabase): address database security advisor warnings

chiptus:
Fixes mutable search_path on 15 `SECURITY DEFINER`/trigger functions (pinned to `''`, the strictest setting, since every body is fully schema-qualified — including `duplicate_set_with_votes`, whose previously-bare `sets`/`set_artists`/`votes` references were qualified to make that safe), tightens always-true RLS policies on `artist_music_genres`, dedupes a redundant public-listing storage policy on `festival-assets`, and revokes unnecessary `SECURITY DEFINER` execute grants — from `PUBLIC` as well as `anon`/`authenticated`, since Postgres grants EXECUTE to PUBLIC by default and both roles inherit it (`bootstrap_super_admin`, `check_username_exists`, `duplicate_set_with_votes` for `anon`).

## Verification
- Review the migration SQL for correctness against current function/policy signatures.
- After applying: core team can still create/edit/delete artist music genres; other authenticated users cannot.
- Festival assets still load via their public URLs (single remaining SELECT policy).
- `bootstrap_super_admin` and `check_username_exists` RPCs are no longer callable by `anon`/`authenticated` clients (still usable via direct DB connection, e.g. `scripts/recreate-staging.sh`).
- `duplicate_set_with_votes` still works end-to-end for `authenticated` callers after the schema-qualification and `search_path=''` change.
- Run `supabase db advisors` (or MCP `get_advisors`) after deploy to confirm the warnings are cleared.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FJkNvMdqnanZ1HBjMN8vdJ)_


supabase/migrations/20260818190000_fix_security_advisor_warnings.sql
```diff
@@ -0,0 +1,135 @@
...
+
+-- Postgres grants EXECUTE to PUBLIC by default on every new function, and
+-- both `anon` and `authenticated` inherit PUBLIC's privileges. Revoking only
+-- from the named roles below would leave them able to call through the
+-- inherited PUBLIC grant, so PUBLIC must be revoked explicitly too.
+
+-- One-time local-dev bootstrap for the first super admin. Already
+-- self-guards (only succeeds while zero super admins exist), but there is
+-- no reason for it to remain a permanently public RPC endpoint.
+REVOKE EXECUTE ON FUNCTION public.bootstrap_super_admin(text) FROM PUBLIC, anon, authenticated;
```
chiptus, line 223: But then it won't be usable on dev/staging, right?

chiptus: It should still work — `check_username_exists` isn't called directly from the frontend (no `.rpc('check_username_exists')` anywhere in `src/`), only from inside `validate_profile_update` and `handle_new_user`, which are both `SECURITY DEFINER`. When a `SECURITY DEFINER` function calls another function internally, that inner call runs with the *definer's* privileges, not the original caller's — so it doesn't need `anon`/`authenticated` to have their own `EXECUTE` grant on `check_username_exists`. That holds the same way on dev/staging/prod, since it's a privilege model thing, not environment-specific.

If there's a dev/staging path I'm missing that calls it directly (a script, a Supabase Studio SQL editor session, etc.), let me know and I'll adjust — but nothing in the repo calls it that way today.

---

chiptus: Correction — I misread which line this was on and answered about `check_username_exists` instead of `bootstrap_super_admin`. Sorry about that. Re-answering for the right function:

`bootstrap_super_admin` should still work on dev/staging: `scripts/recreate-staging.sh` calls it via `psql "$STAGING_DB_URL"` (a direct Postgres connection using the DB owner/service role), not through PostgREST as `anon`/`authenticated`. Direct psql connections as the DB owner bypass `GRANT`/`REVOKE` on `PUBLIC`/`anon`/`authenticated` entirely, so that script is unaffected.

What the revoke *does* change: previously any signed-in (or even anonymous) client could call `POST /rest/v1/rpc/bootstrap_super_admin` directly from the browser. The function self-guards against creating a second super admin, but there's no reason for it to be reachable as a public REST endpoint at all — the only real caller is the staging-recreation script running with DB-owner privileges.

If there's another flow that bootstraps the first admin by having a logged-in browser user call this RPC directly (rather than via the script), that would break and I'd want to leave the `authenticated` grant in place — let me know if that's the case.

---

PR #302, opened 2026-08-18
test(integration): add fixture factories and scratch pool

### LINKED ISSUE #281: Fixture factories + scratch pool for integration tests
## Parent

#64

## What to build

Extend the integration-test harness (#280) with fixture factories so tests can create scoped data without hand-writing inserts and cleanup, and stay safe under parallel test workers.

- Factory functions (e.g. create-artist, create-set, create-user) that insert via the real Supabase client, return the created row's id, and self-register with the harness's shared cleanup hook so nothing needs to be manually deleted.
- Every factory generates a uniquely-named/scoped row (no two tests ever touch the same row), rather than relying on shared mutable seed data.
- A pool of disposable "scratch" rows for common simple cases, so not every test needs to build a full dependency graph from scratch.
- Proven by an integration test for a hook that reads data the test itself creates (e.g. a hook scoped to an edition/artist/set), demonstrating that running it repeatedly or in parallel produces no collisions and leaves no residue.

## Acceptance criteria

- [ ] At least two fixture factories exist (e.g. artist and set), each returning the created row's id and auto-registering for cleanup.
- [ ] A scratch-row pool exists for at least one common case.
- [ ] Running the proving integration test twice back-to-back (or concurrently) produces no collisions and leaves the database in the same state it started in.
- [ ] No test in this ticket hand-writes its own insert/delete cleanup code — everything goes through the factories.

## Blocked by

- #280 (integration-test tier scaffold + harness core)

chiptus:
Adds `createArtist`/`createSet`/`createScratchFestivalEdition`/`linkArtistToSet` fixture factories that insert via the real Supabase client, return the created row's id, and self-register cleanup with the integration harness, so tests never hand-write insert/delete boilerplate.
`createSet` falls back to a fresh scratch festival edition when no `festival_edition_id` is given, so callers needing just *some* valid edition skip the festivals → festival_editions chain.

## Verification
- Start local Supabase (`supabase start`), then run `pnpm test:integration` — `useSetsByEdition.integration.test.ts` passes, building its own edition/artist/set via the factories and asserting `setsByEditionQuery` returns only what it created.
- Run the same test twice back-to-back (or concurrently); no collisions, and the database ends in the same state it started in.
- Run `pnpm test` — the new integration test file is excluded and the existing unit suite (485 tests) passes unaffected.
- Run `pnpm run typecheck` and `pnpm run lint` — both clean.

Closes #281

---
_Generated by [Claude Code](https://claude.ai/code/session_017toUhvwdoZrGk8i9ahtyEX)_


src/api/sets/useSetsByEdition.integration.test.ts
```diff
@@ -0,0 +1,35 @@
...
+import { setsByEditionQuery } from "./useSetsByEdition";
+import { createQueryWrapper } from "@/test/integration/harness";
+import { createArtist } from "@/test/integration/fixtures/artists";
+import { createSet } from "@/test/integration/fixtures/sets";
+import { createScratchFestivalEdition } from "@/test/integration/fixtures/scratchPool";
+import { linkArtistToSet } from "@/test/integration/fixtures/setArtists";
+
+describe("setsByEditionQuery", () => {
+  // Each run creates its own rows, so repeats never collide.
+  it.each(["first run", "second run"])(
```
chiptus, line 13: What are we testing here?

PR #306, opened 2026-08-18
test(auth): add signInAsTestUser and vote mutation reference test

### LINKED ISSUE #282: Auth helper + mutation invalidate/refetch reference test
## Parent

#64

## What to build

An auth helper for RLS-gated integration tests, plus the reference mutation-hook test that proves the harness handles the real commit → invalidate → refetch cycle, not just a mutation's own response.

- `signInAsTestUser()`-style helper: mints a session via `supabase.auth.admin.generateLink({ type: 'magiclink' })` + `verifyOtp()` against the local instance's service-role key. No real email delivery, no polling the local stack's Mailpit inbox.
- A reference integration test for a mutation hook with an invalidate/refetch cycle (e.g. `useVoteMutation`): signs in as a test user, creates the fixture data it needs via the factories from #281, performs the mutation, and asserts the *query* result reflects it after invalidation — not just that the mutation's own response looked right.
- A real RLS-denial error-path test (e.g. attempting the mutation as an unauthorized/unauthenticated user and asserting the real error Supabase/PostgREST returns).
- Written documentation (in the harness module or a short README near it) of the mocking policy: no mocks by default; a narrow, explicit exception permits mocking at the fetch/network boundary — never the hook — reserved for conditions impractical to reproduce for real (e.g. genuine network unreachability). Error conditions that are cheap to trigger for real (RLS denial, unique-constraint violation, not-found) must use the real condition.

## Acceptance criteria

- [ ] `signInAsTestUser()` (or equivalent) returns a real, usable session without waiting on email delivery.
- [ ] The reference mutation test asserts on the refetched query result post-invalidation, using fixture data created via #281's factories.
- [ ] A real RLS-denial test exists and asserts on the actual error Supabase returns.
- [ ] The no-mocks-by-default / fetch-boundary-exception policy is written down somewhere a future contributor will find it.

## Blocked by

- #281 (fixture factories + scratch pool)

chiptus:
Adds a `signInAsTestUser()` fixture that mints a real session via the admin API (no email/Mailpit), exports `vote()` from `useVoteMutation` so tests call it directly, and adds two reference integration tests: an authenticated vote persisting through RLS, and an unauthenticated vote rejected with the real RLS-denial error. Documents the mocking policy in `src/test/integration/README.md`, including guidance against duplicating TanStack Query's own cache mechanics when an existing E2E spec already proves the same outcome.

## Verification
- Start local Supabase (`supabase start`), run `pnpm test:integration` — both `useVoteMutation.integration.test.ts` cases pass: an authenticated vote lands in the database under RLS, and an unauthenticated vote both throws from `vote()` and fails with real Postgres error code `42501`.
- Run `pnpm test` — integration tests stay excluded, existing unit suite (489 tests) passes unaffected.
- Run `pnpm run typecheck` and `pnpm run lint` — both clean.

Closes #282

---
_Generated by [Claude Code](https://claude.ai/code/session_017toUhvwdoZrGk8i9ahtyEX)_

### DIFF
--- a/src/api/voting/useVoteMutation.integration.test.ts
+++ b/src/api/voting/useVoteMutation.integration.test.ts
@@ -1,9 +1,8 @@
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 import { act, renderHook, waitFor } from "@testing-library/react";
 import { useQuery } from "@tanstack/react-query";
 import { useVoteMutation } from "./useVoteMutation";
 import { userVotesQuery } from "./useUserVotesQuery";
-import { userVotesKeys } from "./types";
 import { supabase } from "@/integrations/supabase/client";
 import { createQueryWrapper } from "@/test/integration/harness";
 import { signInAsTestUser } from "@/test/integration/fixtures/auth";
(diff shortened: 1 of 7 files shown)


src/api/voting/useVoteMutation.integration.test.ts
```diff
@@ -0,0 +1,108 @@
...
+import { signInAsTestUser } from "@/test/integration/fixtures/auth";
+import { createSet } from "@/test/integration/fixtures/sets";
+import { SEEDED_USER_ID } from "@/test/integration/fixtures/constants";
+
+describe("useVoteMutation", () => {
+  it("commits through a real invalidate -> refetch cycle, not just the mutation's optimistic write", async () => {
+    const userId = await signInAsTestUser();
+    const setId = await createSet();
+
+    const queryClient = new QueryClient({
```
chiptus: Why don't we use the wrapper helper?

chiptus: Good catch — this test needed a handle on the `QueryClient` itself (to subscribe to `getQueryCache()` and watch `fetchStatus`), which `createQueryWrapper()` didn't expose. Rather than duplicate the setup, I updated the helper to return `{ Wrapper, queryClient }` and switched all call sites (including this one) to use it — pushed in 40ecc77.

---
_Generated by [Claude Code](https://claude.ai/code/session_017toUhvwdoZrGk8i9ahtyEX)_

---

src/api/voting/useVoteMutation.integration.test.ts
```diff
@@ -0,0 +1,93 @@
...
+      { wrapper: Wrapper },
+    );
+
+    await waitFor(() => expect(result.current.votes.isSuccess).toBe(true));
+
+    // Only start watching fetchStatus once the initial mount fetch has
+    // settled, so the assertion below isolates activity the mutation causes
+    // and doesn't just see that first fetch.
+    const fetchStatuses: string[] = [];
+    const unsubscribe = queryClient.getQueryCache().subscribe((event) => {
```
chiptus: Feels like checking implementation details

chiptus: Agreed — replaced it with a spy on the real `queryFn` (still calls through, so the real network fetch happens) and counted invocations instead: 1 after mount, 2 after the mutation settles. That proves a genuine refetch occurred without reaching into TanStack Query's internal `fetchStatus` state, and it also drops the subscription/cleanup Copilot flagged earlier since there's no manual subscription anymore. Pushed in 0e997c1.

---

src/api/voting/useVoteMutation.integration.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    const { error } = await supabase
+      .from("votes")
+      .upsert(
+        { user_id: SEEDED_USER_ID, set_id: setId, vote_type: 1 },
+        { onConflict: "user_id,set_id" },
+      );
+
+    expect(error).not.toBeNull();
+    expect(error?.code).toBe("42501");
+  });
```
chiptus, line 50: but this doesn't test nothing in our api, just supabase and the fixtures

chiptus: Right, it never called `vote()` at all. Fixed: it now asserts `vote()` itself throws on the denial (exercising our error-propagation code), then separately confirms via a raw call that the underlying cause really is RLS — not something else `vote()` would have equally swallowed into its generic Error. Pushed in 73eadae.

---

chiptus:
CI red on `test (1, 4) / run` — firefox-only failure in `tests/e2e/navigation.spec.ts:5:3` (nav click doesn't land on `/schedule`). This predates this PR: the identical failure hit the same shard on this branch ~11 hours ago at commit dee3a48, before any of today's auth/voting changes existed, and this PR's diff doesn't touch routing/navigation code at all. Not fixing here — will re-check when it clears.

---
_Generated by [Claude Code](https://claude.ai/code/session_017toUhvwdoZrGk8i9ahtyEX)_

---

PR #310, opened 2026-08-19
feat(schedule): scope vote-chip filters to Everyone/Me/Group

chiptus:
Schedule tab's vote-type filter chips now respect the navbar's Active Scope switcher (Everyone / Me / Active Group) instead of a separate, Schedule-only Me/Group toggle. This removed the earlier `voteScope` URL param and the extra selector — the chips just read whatever scope is already active globally.

## Verification
- Log in, open Schedule (list or timeline), set the navbar switcher to Me, select a vote-type chip; only your own matching votes filter sets.
- Set the navbar switcher to Active Group; a chip now matches sets where any group member (including you) voted that way, immediately, no reload.
- Set the navbar switcher to Everyone; a chip matches sets where anyone at all voted that way.
- Select multiple chips; OR semantics hold under all three scopes.
- Switching the navbar scope re-filters Schedule's visible sets immediately.

---
_Generated by [Claude Code](https://claude.ai/code/session_012xAfdayZzwNNJ8EGap3Rix)_


chiptus (commented):
Addressed all three review comments in f2d754a: renamed the mobile "My vote" label to scope-neutral "Vote", reused a shared empty Set instead of allocating one per filtered set, and gated the group-members query on the resolved scope being "group".

---

src/hooks/useScheduleVoteScope.ts
```diff
@@ -0,0 +1,56 @@
...
+ * when one was made, else Active Group when the user has one, else Me.
+ * Also resolves the Active Group's member ids for group-scope filtering,
+ * `undefined` while loading (or when there's no active group).
+ */
+export function useScheduleVoteScope(tab: "timeline" | "list") {
+  const { user } = useAuth();
+  const { activeGroupId } = useActiveScope();
+  const { voteScope: urlVoteScope, updateVoteScope } = useTimelineUrlState(tab);
+  const { data: groups } = useQuery({
+    ...userGroupsQuery(user?.id ?? ""),
```
chiptus: Don't we have a fetch group query?

chiptus: No fetch-single-group-by-id query exists — `src/api/groups/` only has `groupBySlugQuery` (needs a slug, not an id) and `userGroupsQuery` (fetches all of a user's groups). This mirrors the exact same pattern `AuthedFilteredSetsPanel.tsx` already uses on the Artists tab to resolve the Active Group's name for `VotePerspectiveToggle` — `useSuspenseQuery(userGroupsQuery(props.userId))` then `.find((group) => group.id === activeGroupId)`. Happy to add a dedicated `groupByIdQuery` if you'd rather have one, but wanted to flag it's consistent with existing code first.

---

chiptus: Moot now — pushed 37a0bf5 removing the separate Schedule Vote Scope toggle entirely per your feedback. The hook no longer resolves a group *name* at all (so no `userGroupsQuery` lookup either); it just reads `useActiveScope().current.kind` directly and fetches `groupMembersQuery` only when that's `"group"`. Vote-chip scope now follows the navbar switcher (Everyone/Me/Group) with no Schedule-local UI or URL state.

---

src/components/layout/AppHeader/GroupSwitcher/ActiveGroupSwitcher.tsx
```diff
@@ -36,6 +36,7 @@ export function ActiveGroupSwitcher({
           variant="outline"
           size={isMobile ? "sm" : "default"}
           className={className}
+          data-testid="active-scope-switcher"
```
chiptus: don't use test id

chiptus: Done in bd0e8d9 — dropped the `data-testid` and made the button's `aria-label` unconditional (previously only set on mobile) so the e2e test can target it by accessible name (`getByRole("button", { name: /^Active scope:/ })`) instead.

---

tests/utils/groups.ts
```diff
@@ -71,3 +71,26 @@ export async function createGroupWithMember(
...
 }
+
+// Adds an existing group's membership for another user.
+export async function addMemberToGroup(
+  groupId: string,
+  email: string,
+): Promise<void> {
+  const userId = await getUserIdByEmail(email);
+
+  const memberResponse = await fetch(
```
chiptus, line 82: why are we using fetch and not supabase client?

chiptus: Matches the existing pattern in this file (`createGroupWithMember`/`getUserIdByEmail` above already use raw `fetch` with the service-role key) — there's no Supabase JS client instantiated anywhere in the test utils (`tests/config/test-env.ts` only exports the service-role key as a string, not a client). These admin-only test setup calls need the service-role key to bypass RLS, which is why they go straight to the REST API rather than through a client instance that doesn't exist here. Happy to introduce a shared `createClient(...)` test helper if you'd like to consolidate this — just didn't want to change the established pattern unprompted in this PR.

---

PR #311, opened 2026-08-19
chore(types): enable exactOptionalPropertyTypes

### LINKED ISSUE #37: Consider enabling exactOptionalPropertyTypes in tsconfig
## Background

`exactOptionalPropertyTypes` is a TypeScript strictness flag. It changes optional properties so that `name?: string` means the key may be *absent* — but you can no longer explicitly assign `undefined` to it.

```ts
type T = { name?: string };
const a: T = { name: undefined }; // OK today, ERROR with the flag
const b: T = {};                  // OK either way
```

This catches the genuine distinction between "key absent" and "key explicitly `undefined`", which matters for `Object.keys()`, spreads, `in` checks, etc.

## Why it's a separate task

- It is repo-wide hardening: enabling it would surface type errors across many files, not a localized change.
- It conflicts with patterns already in the codebase. For example, the mutation payloads in `useUpdateSet.ts` / `useUpdateArtist.ts` build a single object literal that assigns possibly-`undefined` values directly (relying on JSON serialization to drop them). With the flag on, those would need to go back to conditional/guarded construction.

## Suggested approach

1. Enable `exactOptionalPropertyTypes` in `tsconfig.app.json`.
2. Run `pnpm run typecheck`, triage the resulting errors.
3. Fix incrementally, or decide it is not worth the churn.

## Context

Came up during PR #34 review while refactoring set/artist mutation payloads. Tracking separately so the decision is deliberate rather than bundled into an unrelated PR.

chiptus:
Enables `exactOptionalPropertyTypes` in `tsconfig.app.json` and fixes the resulting type errors, per #37. Most fixes widen prop/param types to explicitly accept `undefined`; mutation payload builders (`useUpdateSet`, `useUpdateProfile`, `useCreateGroup`, `useGenerateInviteMutation`, `parseCsv`) now build their objects with conditional spreads instead of assigning `undefined` directly.

Closes #37.

## Verification
- `npx tsc --noEmit -p tsconfig.app.json` — 0 errors
- `pnpm run lint` — 0 errors (2 pre-existing unrelated warnings)
- `pnpm test -- --run` — 494/494 passing
- Manually check the Active Group toggle in Settings and the Stage selector in admin still show "nothing selected" correctly when no value is set (fixed to omit the `value` prop rather than coerce to `""`)

---
_Generated by [Claude Code](https://claude.ai/code/session_01QvzpJZujRJYbrrUg17nT78)_

### DIFF
--- a/src/components/ui/context-menu.tsx
+++ b/src/components/ui/context-menu.tsx
@@ -90,14 +90,13 @@ ContextMenuItem.displayName = ContextMenuPrimitive.Item.displayName;
 const ContextMenuCheckboxItem = React.forwardRef<
   React.ElementRef<typeof ContextMenuPrimitive.CheckboxItem>,
   React.ComponentPropsWithoutRef<typeof ContextMenuPrimitive.CheckboxItem>
->(({ className, children, checked, ...props }, ref) => (
+>(({ className, children, ...props }, ref) => (
   <ContextMenuPrimitive.CheckboxItem
     ref={ref}
     className={cn(
       "relative flex cursor-default select-none items-center rounded-sm py-1.5 pl-8 pr-2 text-sm outline-none focus:bg-accent focus:text-accent-foreground data-[disabled]:pointer-events-none data-[disabled]:opacity-50",
       className,
     )}
-    checked={checked}
     {...props}
   >
     <span className="absolute left-2 flex h-3.5 w-3.5 items-center justify-center">
--- a/src/pages/Settings/ActiveGroupSetting.tsx
+++ b/src/pages/Settings/ActiveGroupSetting.tsx
@@ -24,7 +24,7 @@ export function ActiveGroupSetting({
       </p>
       <ToggleGroup
         type="single"
-        value={activeGroupId}
+        value={activeGroupId ?? ""}
         onValueChange={(value) => {
           if (value) {
             setActiveGroup(value);
--- a/src/pages/Settings/ActiveGroupSetting.tsx
+++ b/src/pages/Settings/ActiveGroupSetting.tsx
@@ -24,7 +24,7 @@ export function ActiveGroupSetting({
       </p>
       <ToggleGroup
         type="single"
-        value={activeGroupId ?? ""}
+        {...(activeGroupId !== undefined ? { value: activeGroupId } : {})}
         onValueChange={(value) => {
           if (value) {
             setActiveGroup(value);
--- a/src/pages/Settings/ActiveGroupSetting.tsx
+++ b/src/pages/Settings/ActiveGroupSetting.tsx
@@ -24,7 +24,7 @@ export function ActiveGroupSetting({
       </p>
       <ToggleGroup
         type="single"
-        {...(activeGroupId !== undefined ? { value: activeGroupId } : {})}
+        value={activeGroupId}
         onValueChange={(value) => {
           if (value) {
             setActiveGroup(value);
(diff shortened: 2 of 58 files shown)


src/components/ui/context-menu.tsx
```diff
@@ -90,14 +90,13 @@ ContextMenuItem.displayName = ContextMenuPrimitive.Item.displayName;
...
   React.ComponentPropsWithoutRef<typeof ContextMenuPrimitive.CheckboxItem>
->(({ className, children, checked, ...props }, ref) => (
+>(({ className, children, ...props }, ref) => (
   <ContextMenuPrimitive.CheckboxItem
     ref={ref}
     className={cn(
       "relative flex cursor-default select-none items-center rounded-sm py-1.5 pl-8 pr-2 text-sm outline-none focus:bg-accent focus:text-accent-foreground data-[disabled]:pointer-events-none data-[disabled]:opacity-50",
       className,
     )}
-    checked={checked}
```
chiptus, line 100: why is this removed?

src/pages/admin/festivals/StageSelector.tsx
```diff
@@ -22,7 +22,11 @@ export function StageSelector({
   return (
     <div className="space-y-2">
       <Label htmlFor="stage">Stage</Label>
-      <Select value={value} onValueChange={onValueChange} disabled={isLoading}>
+      <Select
+        {...(value !== undefined ? { value } : {})}
```
chiptus: can we change select to accept an undefined value?

src/pages/Settings/ActiveGroupSetting.tsx
```diff
@@ -24,7 +24,7 @@ export function ActiveGroupSetting({
       </p>
       <ToggleGroup
         type="single"
-        value={activeGroupId}
+        {...(activeGroupId !== undefined ? { value: activeGroupId } : {})}
```
chiptus: can't we change toggle group to accept undefined?

src/components/ui/context-menu.tsx
chiptus: `checked` is still passed through — it's part of `...props` now instead of being destructured and reassigned explicitly. Under `exactOptionalPropertyTypes`, Radix's `checked` prop is typed `checked?: boolean | "indeterminate"` (no explicit `| undefined`), so reconstructing `checked={checked}` from a local const whose inferred type includes `undefined` fails the same way `value={value}` did elsewhere in this PR. Since `checked` already flows through untouched via `{...props}`, destructuring it out just to hand it straight back was redundant, so I dropped the explicit reassignment rather than working around it.

---

src/pages/admin/festivals/StageSelector.tsx
chiptus: Yes — pushed in d877e1d. `Select` in `src/components/ui/select.tsx` now wraps `SelectPrimitive.Root` with a widened `value?: string | undefined` prop and omits the key internally when it's undefined, instead of every call site doing that. `StageSelector.tsx` now just passes `value={value}` directly again. (We can't widen Radix's own type — it's in `node_modules` and declares `value?: string` with no `undefined` — so the workaround has to live in our local wrapper.)

---

src/pages/Settings/ActiveGroupSetting.tsx
chiptus: Same fix as StageSelector, pushed in d877e1d — `ToggleGroup` in `src/components/ui/toggle-group.tsx` now widens `value` (preserving the single/multiple discriminated union) and omits it internally when undefined. `ActiveGroupSetting.tsx` passes `value={activeGroupId}` directly again.

---

PR #328, opened 2026-08-20
feat(invites): add dedicated /invite route

chiptus:
Replaces the split invite handling in `__root.tsx` (inline overlay for anonymous users, silent toast for authed ones) with one dedicated `/invite` screen both states pass through. Root now redirects any request carrying `?invite=` to `/invite`, remembering where the user was actually headed so they land back there after accepting — same host, since it's all client-side navigation. Authed accept failures now get a real error screen with retry instead of a toast. Also retires the two single-item auto-redirect guards from the earlier fix, since the root-level redirect now intercepts before they'd ever see a pending invite.

## Verification
- Open an invite link while logged out; lands on `/invite` and shows the "you're invited" screen, unaffected by which route the link originally pointed to.
- Open an invite link while logged in; shows a "Joining…" screen, then redirects to the original destination on success.
- Trigger an accept failure (e.g. expired/overused invite while authed); shows an error screen with a working "Try again" button.
- Open an invite link on a subdomain whose festival has exactly one edition; still processes correctly (guard removal covered).
- Open `/invite?invite=TOKEN` directly with no `redirect` param; lands on `/` after accepting.
- Open an invite link on the main domain and a subdomain with multiple editions; both still work as before.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FpiBhZGqy114fNQKqj81oW)_


src/routes/invite.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const navigate = useNavigate();
+
+  const inviteQuery = useInviteValidationQuery(inviteToken);
+  const {
+    mutate: acceptInvite,
+    isPending: isAccepting,
+    isError: acceptFailed,
+    error: acceptError,
+    reset: resetAccept,
+  } = useAcceptInviteMutation();
```
chiptus: why do we destructure? just `const acceptMutation = use...`

src/routes/invite.tsx
```diff
@@ -0,0 +1,165 @@
...
+        inviteValidation={inviteQuery.data}
+        inviteToken={inviteToken}
+      />
+    );
+  }
+
+  return null;
+}
+
+function StatusScreen({ children }: { children: React.ReactNode }) {
```
chiptus:
```suggestion
function StatusScreen({ children }: { children: ReactNode }) {
```

src/routes/invite.tsx
```diff
@@ -0,0 +1,165 @@
+import { createFileRoute, useNavigate } from "@tanstack/react-router";
```
chiptus, line 1: decompose component, it will make it simpler

src/components/invite/useAcceptInviteFlow.ts
```diff
@@ -0,0 +1,58 @@
...
+      );
+    },
+    [inviteToken, acceptMutation, navigate, redirect, toast, inviteQuery.data],
+  );
+
+  useEffect(() => {
+    const isValid = inviteQuery.data?.is_valid === true;
+    if (
+      !userId ||
+      isValid !== true ||
```
chiptus:
```suggestion
      !isValid ||
```

